PR #35, opened 2024-10-04
[SL-Temp]EM requirement removal at AppInit

lpbeliveau-silabs:
Added EM req requirement removal to AppInit temporarily until ot patch


mkardous-silabs (changes requested):
This brakes gn builds that use sdk_support since the fix is already present there.

lpbeliveau-silabs:
Added ifdef check for gn build

jmartinez-silabs (approved):

lpbeliveau-silabs:
merging on stash matter_sdk so no need for double PR

PR #56, opened 2024-10-17
[SL-UP] Update WiseConnect to v3.3.3

arun-silabs:
This PR contains the changes required for updating WiseConnect to 3.3.3.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -223,7 +223,7 @@
 [submodule "third_party/silabs/wifi_sdk"]
 	path = third_party/silabs/wifi_sdk
 	url = https://github.com/SiliconLabs/wiseconnect.git
-	branch = v3.3.2
+	branch = v3.3.3
 	platforms = silabs
 [submodule "editline"]
 	path = third_party/editline/repo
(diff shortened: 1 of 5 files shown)


.gitmodules
```diff
@@ -223,7 +223,7 @@
 [submodule "third_party/silabs/wifi_sdk"]
 	path = third_party/silabs/wifi_sdk
 	url = https://github.com/SiliconLabs/wiseconnect.git
-	branch = v3.3.2
```
mykrupp, line 226: This isn't necessary to do in matter_sdk, if we want to do this update it should be done in CSA. We use the wifi_sdk from the matter_extension, so we can still comeplete the task to move to 3.3.3 by updating matter_extensions wifi_sdk submodule.

third_party/silabs/matter_support
mykrupp, line 1: Same comment on this, these should only be updated in CSA.

third_party/silabs/wifi_sdk
mykrupp, line 1: Same comment on this, these should only be updated in CSA.

mkardous-silabs (changes requested):
@arun-silabs We can still cherry-pick changes to th  matter v1.4 branch. Instead of a cherry-pick PR, please open a PR to port these changes to CSA v1.4 branch once the PR towards master is merged.

PR #66, opened 2024-10-22
[SL-TEMP]Add Matter+zigbee cmp logic On matter commissioning complete

jmartinez-silabs:
Fixes [MATTER-4273](https://jira.silabs.com/browse/MATTER-4273)

Add a snippet of code that was missing from the older branch structure to handle the Zigbee network when matter commissioning completes.

_Using [sl-temp] tag in the release branch as I will cherry-pick to Main immediately_

### DIFF
--- a/examples/platform/silabs/BaseApplication.cpp
+++ b/examples/platform/silabs/BaseApplication.cpp
@@ -74,6 +74,13 @@
 #include <performance_test_commands.h>
 #endif // PERFORMANCE_TEST_ENABLED
 
+// SL-Only
+#include "sl_component_catalog.h"
+#ifdef SL_CATALOG_ZIGBEE_STACK_COMMON_PRESENT
+#include "ZigbeeCallbacks.h"
+#include "sl_cmp_config.h"
+#endif
+
 /**********************************************************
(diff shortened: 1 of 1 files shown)


mkardous-silabs (approved):

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -74,6 +74,13 @@
 #include <performance_test_commands.h>
 #endif // PERFORMANCE_TEST_ENABLED
 
+// SL-Only
+#include "sl_component_catalog.h"
```
mykrupp, line 78: Just wanted to double check that you are intentionally including "sl_component_catalog.h" for all apps here?

mykrupp (approved):

examples/platform/silabs/BaseApplication.cpp
jmartinez-silabs: Yes, I want to use `SL_CATALOG_ZIGBEE_STACK_COMMON_PRESENT` and the include only brings existing definitions to the BaseApplication.cpp. No consequence for other apps

PR #76, opened 2024-10-26
[SL-UP] Skip the OpenThread API call for WiFi builds

selissia:
Build fails for thermostat-rs911x.slcp if the matter_shell component is excluded. To fix skip the OpenThread API call in uart.cpp in case of WiFi builds. 
Verified in a FULL CI build on the matter_extension repo.

### DIFF
--- a/examples/platform/silabs/uart.cpp
+++ b/examples/platform/silabs/uart.cpp
@@ -413,7 +413,7 @@ static void UART_rx_callback(UARTDRV_Handle_t handle, Ecode_t transferStatus, ui
 
 #ifdef ENABLE_CHIP_SHELL
     chip::NotifyShellProcess();
-#elif !defined(PW_RPC_ENABLED)
+#elif !defined(PW_RPC_ENABLED) && !defined(SL_WIFI)
     otSysEventSignalPending();
 #endif
 }


senthilku (approved):

examples/platform/silabs/uart.cpp
```diff
@@ -413,7 +413,7 @@ static void UART_rx_callback(UARTDRV_Handle_t handle, Ecode_t transferStatus, ui
 
 #ifdef ENABLE_CHIP_SHELL
     chip::NotifyShellProcess();
-#elif !defined(PW_RPC_ENABLED)
+#elif !defined(PW_RPC_ENABLED) && !defined(SL_WIFI)
```
mkardous-silabs, line 416: Should we instead define gate with `CHIP_DEVICE_CONFIG_THREAD_ENABLE_CLI` since the Ot-Cli should not be enabled?

selissia: Perhaps. But not in this PR which is a hotfix to unblock the release build. The SL_WIFI macro is already used in other places in this file.

mkardous-silabs: I don't see the relevance with a hotfix but i won't block the PR. Can you file a follow up item when the upstream PR is open?

mkardous-silabs (approved):

examples/platform/silabs/uart.cpp
selissia: MATTER-4295 has been created to follow up

PR #78, opened 2024-10-29
[SL-UP] Split unique id and persistent unique

jmartinez-silabs:
The Latest Specification made it clear that the Unique ID used for the rotation Device ID and the UniqueID attribute from the basic information cluster are not the same.

Our previously provisioned UniqueID remains untouched but it is now renamed PersitentUniqueID to make it clear it is not the same as The UniqueId Attribute as presented in spec.

The UniqueID attribute from the basicInformation value is now stored in our NVM3 config section. It is erased on factory reset. After factory reset, at init, we will generate a random unique id. A user can store a New UniqueID at runtime if he decides to with StoreUniqueId()
This code already exists so the only change in this PR on this topic is having the kConfigKey_UniqueId in the NVM3 config section which is erased on factory reset

Also ReturnErrorCodeIf doesn't exist anymore so replaced with VerifyOrReturnError

### DIFF
--- a/examples/platform/silabs/provision/ProvisionStorageDefault.cpp
+++ b/examples/platform/silabs/provision/ProvisionStorageDefault.cpp
@@ -403,14 +403,14 @@ CHIP_ERROR Storage::GetManufacturingDate(uint8_t * value, size_t max, size_t & s
     return SilabsConfig::ReadConfigValueStr(SilabsConfig::kConfigKey_ManufacturingDate, (char *) value, max, size);
 }
 
-CHIP_ERROR Storage::SetUniqueId(const uint8_t * value, size_t size)
+CHIP_ERROR Storage::SetPersistentUniqueId(const uint8_t * value, size_t size)
 {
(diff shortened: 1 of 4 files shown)


mykrupp:
Is there any consideration for a user that has a project where they use kConfigKey_UniqueId to retrieve a rotating id key at (kMatterFactory_KeyBase, 0x1F) ? Now kConfigKey_UniqueId would be pointing to a different location in nvm for them.

rcasallas-silabs (approved):

lpbeliveau-silabs (approved):

mykrupp:
Makes sense, the nvm3 location of the key stays the same. I was concerned about the actual key name. 

User has a project that relies on kConfigKey_PersistentUniqueId to retrieve rotating id. Now they get this update and the same function name retrieves from a different location in nvm. 

So yeah I think we just need a release note that highlights the rename.

mykrupp (approved):

PR #79, opened 2024-10-29
[SL-UP] 917 static variable retention fix

lpbeliveau-silabs:
Cherry-pick fix for static variable retention in 917 doorlock app.

### DIFF
--- a/examples/lock-app/silabs/include/LockManager.h
+++ b/examples/lock-app/silabs/include/LockManager.h
@@ -199,10 +199,12 @@ public:
 private:
     struct UnlatchContext
     {
+        static constexpr uint8_t kMaxPinLength = UINT8_MAX;
+        uint8_t mPinBuffer[kMaxPinLength];
+        uint8_t mPinLength;
         chip::EndpointId mEndpointId;
         Nullable<chip::FabricIndex> mFabricIdx;
         Nullable<chip::NodeId> mNodeId;
-        Optional<chip::ByteSpan> mPin;
         OperationErrorEnum mErr;
 
         void Update(chip::EndpointId endpointId, const Nullable<chip::FabricIndex> & fabricIdx,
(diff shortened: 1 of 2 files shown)


examples/lock-app/silabs/include/LockManager.h
```diff
@@ -199,10 +199,12 @@ class LockManager
 private:
     struct UnlatchContext
     {
+        static constexpr uint8_t kMaxPinLenght = UINT8_MAX;
```
mkardous-silabs, line 202: should this be a size_t like the mPinLength member?

examples/lock-app/silabs/include/LockManager.h
```diff
@@ -199,10 +199,12 @@ class LockManager
...
+        uint8_t mPinBuffer[kMaxPinLenght];
```
mkardous-silabs, line 203: why the switch to an array instead of a bytespan?

lpbeliveau-silabs: Bytespan doesn't own its memory.

jmartinez-silabs:
Cherry-pick from where?

examples/lock-app/silabs/include/LockManager.h
lpbeliveau-silabs: Three reasons:

1.  size_t is 32 bits, so it takes more space.
2. I don't really see a need for more than 255 characters to a doorlock pin for our sample app.
3. Since it's in our app code, a client that has a setup with a longer pin can change that part of the code somewhat easily.

jmartinez-silabs (approved):

mkardous-silabs (approved):

examples/lock-app/silabs/include/LockManager.h
mkardous-silabs: in that case, should  mPinLength be of the same type of this? The max is a uint8 but the variable that stores the information is a uint32?

examples/lock-app/silabs/include/LockManager.h
mkardous-silabs: True - i forgot about that

examples/lock-app/silabs/include/LockManager.h
mykrupp, line 202:
```suggestion
        static constexpr uint8_t kMaxPinLength = UINT8_MAX;
```

lpbeliveau-silabs: indeed

lpbeliveau-silabs:
> Cherry-pick from where?

From [slc_1.3](https://stash.silabs.com/projects/WMN_TOOLS/repos/matter/commits) commit [f5c1c55e314](https://stash.silabs.com/projects/WMN_TOOLS/repos/matter/commits/f5c1c55e3143b72b17f25a1606f7233133d5e7af)

examples/lock-app/silabs/include/LockManager.h
lpbeliveau-silabs: changed to uint8_t

PR #81, opened 2024-10-30
[SL-CP] Fixing the scan for the WiFi devices

chirag-silabs:
The scan was broken due to refactor causing issues with ecosystems.

Fixing the scan. Tested on BRD4338A


jmartinez-silabs:
is this a SL-TEMP or a CP-CSA ? If it is a SL-TEMP why is it a TEMP?

examples/platform/silabs/SiWx917/SiWx917/sl_wifi_if.cpp
```diff
@@ -542,14 +542,11 @@ sl_status_t show_scan_results(sl_wifi_scan_result_t * scan_result)
     for (int idx = 0; idx < (int) scan_result->scan_count; idx++)
     {
         memset(&cur_scan_result, 0, sizeof(cur_scan_result));
-
-        cur_scan_result.ssid_length = strnlen((char *) scan_result->scan_info[idx].ssid,
-                                              chip::min<size_t>(sizeof(scan_result->scan_info[idx].ssid), WFX_MAX_SSID_LENGTH));
+        cur_scan_result.ssid_length = chip::min<size_t>(strnlen((char *) scan_result->scan_info[idx].ssid, chip::DeviceLayer::Internal::kMaxWiFiSSIDLength) + 1, chip::DeviceLayer::Internal::kMaxWiFiSSIDLength); // +1 for null termination
```
jmartinez-silabs, line 545: Why are we adding a +1 to the length to accommodate a null termination. That is not how Lenght of a string work.
The fact that you have a length is exactly so you don't have to use strlen and risk doing out-of-memory reads.

There seems to be a some flaws here since it is repeated in every change

examples/platform/silabs/SiWx917/SiWx917/sl_wifi_if.cpp
```diff
@@ -604,7 +601,7 @@ static void wfx_rsi_save_ap_info(void) // translation
     sl_wifi_ssid_t ssid_arg;
     memset(&ssid_arg, 0, sizeof(ssid_arg));
     ssid_arg.length = wfx_rsi.sec.ssid_length;
-    chip::Platform::CopyString((char *) &ssid_arg.value[0], ssid_arg.length, wfx_rsi.sec.ssid);
+    memcpy((char *) &ssid_arg.value[0], wfx_rsi.sec.ssid, ssid_arg.length);
```
jmartinez-silabs, line 604: why are we replacing CopyString by memcpy? Copy string adds some verification/safe checks

examples/platform/silabs/SiWx917/SiWx917/sl_wifi_if.cpp
chirag-silabs: Updated the PR

examples/platform/silabs/SiWx917/SiWx917/sl_wifi_if.cpp
chirag-silabs: Updated the PR

examples/platform/silabs/SiWx917/SiWx917/sl_wifi_if.cpp
```diff
@@ -545,11 +545,11 @@ sl_status_t show_scan_results(sl_wifi_scan_result_t * scan_result)
 
         cur_scan_result.ssid_length = strnlen((char *) scan_result->scan_info[idx].ssid,
                                               chip::min<size_t>(sizeof(scan_result->scan_info[idx].ssid), WFX_MAX_SSID_LENGTH));
-        chip::Platform::CopyString(cur_scan_result.ssid, cur_scan_result.ssid_length, (char *) scan_result->scan_info[idx].ssid);
+        chip::Platform::CopyString(cur_scan_result.ssid, cur_scan_result.ssid_length + 1, (char *) scan_result->scan_info[idx].ssid); // +1 for null termination
```
jmartinez-silabs, line 548: So I didn't notice in my first review but here and in the other places, what is the ssid destination buffer size? From what I can see in wfx_wifi_scan_result struct, ssid buffer is
`char ssid[WFX_MAX_SSID_LENGTH + 1];` if this is always the case then we are good and this is safe.

but if some ssid destination buffer are of size `WFX_MAX_SSID_LENGTH`, and if
`cur_scan_result.ssid_length` is `WFX_MAX_SSID_LENGTH` because of the chip::min at line 546.

the +1 would cause the null termination to be written out of bounds.

Here we seem safe. I'll suggest a comment addition to let other working on that file that the null space is account for

jmartinez-silabs, line 548:
```suggestion
        // cur_scan_result.ssid is of size WFX_MAX_SSID_LENGTH+1, we are safe with the cur_scan_result.ssid_length calculated above
        chip::Platform::CopyString(cur_scan_result.ssid, cur_scan_result.ssid_length + 1, (char *) scan_result->scan_info[idx].ssid); // +1 for null termination
```

examples/platform/silabs/SiWx917/SiWx917/sl_wifi_if.cpp
```diff
@@ -604,7 +604,7 @@ static void wfx_rsi_save_ap_info(void) // translation
     sl_wifi_ssid_t ssid_arg;
     memset(&ssid_arg, 0, sizeof(ssid_arg));
     ssid_arg.length = wfx_rsi.sec.ssid_length;
```
jmartinez-silabs, line 606: for `sl_wifi_ssid_t` ` .value` buffer is of size 32 while `wfx_rsi.sec `is a `wfx_wifi_provision_t` struct where `wfx_rsi.sec.ssid` is of size 33 (WFX_MAX_SSID_LENGTH+1) . we have a risk of out of bound write with the following copystring.

can sl_wifi_ssid_t .value be increased to (WFX_MAX_SSID_LENGTH+1). Can we add a check that the resulting ssid_arg.length is at max size of`  ssid_arg.value - 1`

examples/platform/silabs/efr32/rs911x/rsi_if.cpp
```diff
@@ -757,7 +757,7 @@ void ProcessEvent(WfxEvent_t inEvent)
             memset(&ap, 0, sizeof(ap));
             ap.ssid_length =
                 strnlen(reinterpret_cast<char *>(scan->ssid), chip::min<size_t>(sizeof(scan->ssid), WFX_MAX_SSID_LENGTH));
-            chip::Platform::CopyString(ap.ssid, ap.ssid_length, reinterpret_cast<char *>(scan->ssid));
+            chip::Platform::CopyString(ap.ssid, ap.ssid_length + 1, reinterpret_cast<char *>(scan->ssid)); // +1 for null termination
```
jmartinez-silabs, line 760: here we are safe because of ap.ssid is of size 33 (WFX_MAX_SSID_LENGTH+1) and we do the correct chip::min in the strnlen above. However, I want to point out that `scan->ssid` is of size 34. a new ssid buffer size.

So fare we have ssi buffers of 32, 33 and 34 bytes getting copied around. Lots of risk that we need to catch
```suggestion
            // ap.ssid is of size WFX_MAX_SSID_LENGTH+1, we are safe with the ap.ssid_length calculated above
            chip::Platform::CopyString(ap.ssid, ap.ssid_length + 1, reinterpret_cast<char *>(scan->ssid)); // +1 for null termination
```

examples/platform/silabs/efr32/wf200/host_if.cpp
```diff
@@ -737,7 +737,7 @@ int32_t wfx_get_ap_info(wfx_wifi_scan_result_t * ap)
     int32_t signal_strength;
 
     ap->ssid_length = strnlen(ap_info.ssid, chip::min<size_t>(sizeof(ap_info.ssid), WFX_MAX_SSID_LENGTH));
-    chip::Platform::CopyString(ap->ssid, ap->ssid_length, ap_info.ssid);
+    chip::Platform::CopyString(ap->ssid, ap->ssid_length + 1, ap_info.ssid); // +1 for null termination
```
jmartinez-silabs, line 740:
```suggestion
    // ap->ssid is of size WFX_MAX_SSID_LENGTH+1, we are safe with the ap->ssid_length calculated above
    chip::Platform::CopyString(ap->ssid, ap->ssid_length + 1, ap_info.ssid); // +1 for null termination
```

examples/platform/silabs/SiWx917/SiWx917/sl_wifi_if.cpp
chirag-silabs: @jmartinez-silabs sl_wifi_ssid_t struct value is in wifi sdk and it cannot be modified. 
I was kinda thinking to add a check for time being, something like
`ssid_arg.length = chip::min<size_t>(wfx_rsi.sec.ssid_length, WFX_MAX_SSID_LENGTH - 1);`
This will ensure the copy string will of ssid_arg.value never overflow i.e., get 32 max
`chip::Platform::CopyString((char *) &ssid_arg.value[0], ssid_arg.length + 1, wfx_rsi.sec.ssid); // +1 for null termination`
Once it gets fixed in the wifi sdk then this check can be removed

jmartinez-silabs: since `ssid_arg.value` is not linked in any way to `WFX_MAX_SSID_LENGTH` and it is unrelatedly set to 32, this is situationally true. If `WFX_MAX_SSID_LENGTH` value changes this becomes invalid and it will silently break at an unknown time.

the correct approach would be 
`ssid_arg.length = chip::min<size_t>(wfx_rsi.sec.ssid_length, sizeof(ssid_arg.value) - 1)`

chirag-silabs: Makes sense, updated the PR. Thanks

jmartinez-silabs (approved):

mkardous-silabs (approved):

PR #86, opened 2024-10-31
[SL-UP]Update our refrigerator app zap configs. Enable identify, set default…

jmartinez-silabs:
fixes MATTER-4290 and MATTER-4280

Zap config updates
- Enable Identify cluster on ep 0
- Set default Ref alarm mask attribute to 1. It should start at 1 and user action on the refrigerator can disable it
- Since we already have 2 distinct zap files for Thread and Wifi platforms for this app, make the network commissioning cluster spec-compliant. Enable `SupportedWifiBands` or  `ThreadVersion` and `SupportedThreadFeatures` respectively of the network type.
- Remove Diagnostic log. We do not support it

### DIFF
--- a/examples/refrigerator-app/silabs/data_model/refrigerator-thread-app.zap
+++ b/examples/refrigerator-app/silabs/data_model/refrigerator-thread-app.zap
@@ -41,16 +41,14 @@
         "code": 22,
         "profileId": 259,
         "label": "MA-rootdevice",
-        "name": "MA-rootdevice",
-        "deviceTypeOrder": 0
+        "name": "MA-rootdevice"
       },
       "deviceTypes": [
         {
           "code": 22,
           "profileId": 259,
           "label": "MA-rootdevice",
-          "name": "MA-rootdevice",
-          "deviceTypeOrder": 0
+          "name": "MA-rootdevice"
         }
       ],
       "deviceVersions": [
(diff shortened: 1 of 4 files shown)


mkardous-silabs (approved):

examples/refrigerator-app/silabs/data_model/refrigerator-thread-app.zap
```diff
@@ -41,16 +41,14 @@
...
-        "deviceTypeOrder": 0
+        "name": "MA-rootdevice"
       },
       "deviceTypes": [
         {
           "code": 22,
           "profileId": 259,
           "label": "MA-rootdevice",
-          "name": "MA-rootdevice",
-          "deviceTypeOrder": 0
```
lpbeliveau-silabs, line 53: I had this appear randomly on one of my zap regens, any idea where the `deviceTypeOrder ` comes from? I see we are removing it here, but some zap parameters seem to believe it needs to be added.

lpbeliveau-silabs (approved):

examples/refrigerator-app/silabs/data_model/refrigerator-thread-app.zap
jmartinez-silabs: i don't know to be honest. We could ask the zap team. maybe it as to do with the zap version I have. it is not the latest

PR #94, opened 2024-11-11
[SL-UP] Add OTA Support for series3

jmartinez-silabs:
The implementation of the Series 3 bootloader OTA API differs from that of Series 2. It utilizes an RTOS thread-safe mechanism, so it can't be wrapped in critical sections like Series 2.

Currently, the Secure element is not thread safe in series3, so, during bootloader ota API calls, the thread stack must be locked. 

Also, at the moment, Some OTA process steps on series 3 are slow, and leave the device unresponsive to other task. Add some traces and small osDelay to provide some form of feedback to the user that the device is "working". This is temporary.

Dev notes:
For CMP, Zigbee must be locked to but we don't have this possibility currently.
The Secure element thread safeness will probably be fixed before Zigbee provide us with a locking mechanism so we will see for release.


src/platform/silabs/efr32/OTAImageProcessorImpl.cpp
```diff
@@ -30,6 +30,20 @@ extern "C" {
...
 #include <platform/silabs/SilabsConfig.h>
 
+#ifdef _SILICON_LABS_32B_SERIES_2
+// Series 2 bootloader_ api calls must be called from a critical section context for thread safeness
+#define WRAP_BL_DFU_CALL(code)                                                                                                     \
+    {                                                                                                                              \
+        CORE_CRITICAL_SECTION(code;)                                                                                               \
+    }
+#else
+// series 3 bootloader_ calls use rtos mutex for thread safeness. Cannot be callsed withing a critical sectionß
```
mkardous-silabs, line 40:
```suggestion
// series 3 bootloader_ calls uses rtos mutex for thread safety. Cannot be called within a critical section
```

src/platform/silabs/efr32/OTAImageProcessorImpl.cpp
```diff
@@ -208,11 +231,27 @@ void OTAImageProcessorImpl::HandleFinalize(intptr_t context)
     ChipLogProgress(SoftwareUpdate, "OTA image downloaded successfully");
 }
 
+// TODO SE access is not thread safe. Assert if Other task acess it during bootloader_verifyImage or bootloader_setImageToBootload
```
mkardous-silabs, line 234:
```suggestion
// TODO: SE access is not thread safe. Assert if other tasks accesses it during bootloader_verifyImage or bootloader_setImageToBootload
```

src/platform/silabs/efr32/OTAImageProcessorImpl.cpp
```diff
@@ -235,13 +278,14 @@ void OTAImageProcessorImpl::HandleApply(intptr_t context)
...
         {
             ChipLogError(SoftwareUpdate, "sl_wfx_host_post_bootloader_spi_transfer() error: %ld", err);
-            return;
         }
 #endif // SL_BTLCTRL_MUX
         return;
     }
-
-    CORE_CRITICAL_SECTION(err = bootloader_setImageToBootload(mSlotId);)
+    ChipLogError(SoftwareUpdate, "Image verified, Set image to bootloader");
```
mkardous-silabs, line 285:
```suggestion
    ChipLogProgress(SoftwareUpdate, "Image verified, Set image to bootloader");
```

src/platform/silabs/efr32/OTAImageProcessorImpl.cpp
```diff
@@ -266,8 +309,13 @@ void OTAImageProcessorImpl::HandleApply(intptr_t context)
         return;
     }
 #endif // SL_BTLCTRL_MUX
+
+    ChipLogError(SoftwareUpdate, "Reboot and install new image...");
```
mkardous-silabs, line 313: same comment

src/platform/silabs/efr32/OTAImageProcessorImpl.cpp
```diff
@@ -224,7 +263,11 @@ void OTAImageProcessorImpl::HandleApply(intptr_t context)
         return;
     }
 #endif // SL_BTLCTRL_MUX
-    CORE_CRITICAL_SECTION(err = bootloader_verifyImage(mSlotId, NULL);)
+
+    osDelay(100); // sl-temp: delay for uart print before verifyImage
```
mkardous-silabs, line 267: Is it worth while to have a configurable define for the entire file?

mkardous-silabs (approved):

src/platform/silabs/efr32/OTAImageProcessorImpl.cpp
selissia: Better yet "set image to bootload".

src/platform/silabs/efr32/OTAImageProcessorImpl.cpp
```diff
@@ -30,6 +30,20 @@ extern "C" {
```
selissia, line 30: If/when we decide to support multi-ota for Series 3 the same changes would need to be applied to src/platform/silabs/multi-ota/OTAFirmwareProcessor.cpp

selissia (approved):

src/platform/silabs/efr32/OTAImageProcessorImpl.cpp
jmartinez-silabs: @selissia Yes that can be done in a subsequent pr. The series 3 bootloader and internal image update are still in flux. Once stabilized, we can apply the final state to the muti-ota

PR #97, opened 2024-11-12
[SL-TEMP] Fix for access point power cycle rejoin issue with ecosystem

senthilku:
- The unnecessary join events have been removed. The join event is posted multiple times, causing issues with the ecosystem during AP power cycles. 
- The join retry limit has been removed. During the commissioning process, the MAX_JOIN_RETRIES_COUNT is set to 5, and this limit should be removed once commissioning is successful

Fixes https://jira.silabs.com/browse/MATTER-4297


chirag-silabs (approved):

mkardous-silabs:
Can this PR be closed since a copy of it was opened?

PR #100, opened 2024-11-13
Update main from csa::master

mkardous-silabs:
#### Description
PR updates the main branch from csa::master. A branch was created because their was a conflict to fix in the sl_wifi_interface file where a fix was upstreamed to csa.

#### Tests
CI

PR #102, opened 2024-11-13
Update main from csa::main

mkardous-silabs:
Description
PR updates the main branch from csa::master. A branch was created because their was a conflict to fix in the sl_wifi_interface file where a fix was upstreamed to csa.

Tests
CI


mkardous-silabs:
Merging this PR without review since #100 was merged with the wrong merge type but with reviews.

PR #103, opened 2024-11-14
[SL-UP] Remove two algo design for the wifi retry mechanism

mkardous-silabs:
#### Description
Wi-Fi reconnect logi had two different methodologies : One during commissioning and one for standard operations. Having this double logic is unecessary and the fibonnaci like design is sufficient.

The commissioning design also required access to the BaseApplication which is in the app layer which  is not accessible from the platform layer.

There will be a follow up PR to implement a true fibonnaci retry mechanims since the current design just doubles the retry interval


jmartinez-silabs (approved):

jmartinez-silabs:
Careful with the upstream. the restyle part bring changes to the retrieve script.

scripts/tools/silabs/retrieve_sl_commits.py
```diff
@@ -41,7 +43,7 @@ def main():
     [PREFIX] commits:
     <full_commit_sha> -- <Commit_Title>
     """,
-    epilog= """
+                                     epilog="""
```
rcasallas-silabs, line 46: I think this is a typo

rcasallas-silabs (approved):

lpbeliveau-silabs (approved):
Approve with the condition to look into the typo fix Ricardo pointed out.

scripts/tools/silabs/retrieve_sl_commits.py
mkardous-silabs: @rcasallas-silabs @lpbeliveau-silabs It's not a typo - at least for the restyler.
epilog is an function argument so the restyler alligned it with the function arguemnts

rcasallas-silabs: Restyler does what restyler does...

PR #109, opened 2024-11-15
[SL-UP] Delete unnecessary double abstraction for the wiseconnect power save functions

mkardous-silabs:
#### Description
Wiseconnect wifi interface has a double abstraction :
- wfx_power_save
- wfx_rsi_power_save

Both function executed the same code. As such, removed the wfx_rsi_power_save function to only use the "standard" API.

### DIFF
--- a/src/platform/silabs/wifi/SiWx/WifiInterface.cpp
+++ b/src/platform/silabs/wifi/SiWx/WifiInterface.cpp
@@ -545,37 +545,6 @@ int32_t sl_wifi_platform_disconnect(void)
     return sl_net_down((sl_net_interface_t) SL_NET_WIFI_CLIENT_INTERFACE);
 }
 
-/******************************************************************
- * @fn   wfx_rsi_power_save(rsi_power_save_profile_mode_t sl_si91x_ble_state, sl_si91x_performance_profile_t sl_si91x_wifi_state)
- * @brief
- *       Setting the RS911x in DTIM sleep based mode
- *
- * @param[in] sl_si91x_ble_state : State to set for the BLE
- * @param[in] sl_si91x_wifi_state : State to set for the WiFi
- * @return
- *        None
- *********************************************************************/
-int32_t wfx_rsi_power_save(rsi_power_save_profile_mode_t sl_si91x_ble_state, sl_si91x_performance_profile_t sl_si91x_wifi_state)
-{
-    int32_t status;
-
-    status = rsi_bt_power_save_profile(sl_si91x_ble_state, 0);
-    if (status != RSI_SUCCESS)
-    {
-        ChipLogError(DeviceLayer, "rsi_bt_power_save_profile failed: 0x%lx", static_cast<uint32_t>(status));
-        return status;
-    }
-    sl_wifi_performance_profile_t wifi_profile = { .profile = sl_si91x_wifi_state };
-    status                                     = sl_wifi_set_performance_profile(&wifi_profile);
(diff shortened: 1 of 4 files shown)


src/platform/silabs/wifi/rs911x/WifiInterface.cpp
```diff
@@ -966,8 +933,26 @@ int32_t wfx_rsi_send_data(void * p, uint16_t len)
  * @return  SL_STATUS_OK if successful,
  *          SL_STATUS_FAIL otherwise
  ***********************************************************************/
-sl_status_t wfx_power_save(void) // TODO : Figure out why the extern C is necessary
```
lpbeliveau-silabs, line 969: I take it we figured it out?

mkardous-silabs: yep! i forgot to remove the comment in CSA when i fixed the problem.

src/platform/silabs/wifi/SiWx/WifiInterface.cpp
```diff
@@ -894,9 +863,23 @@ void wfx_dhcp_got_ipv4(uint32_t ip)
  * @return  SL_STATUS_OK if successful,
  *          SL_STATUS_FAIL otherwise
  ***********************************************************************/
-sl_status_t wfx_power_save(rsi_power_save_profile_mode_t sl_si91x_ble_state,
-                           sl_si91x_performance_profile_t sl_si91x_wifi_state) // TODO : Figure out why the extern C is necessary
+sl_status_t wfx_power_save(rsi_power_save_profile_mode_t sl_si91x_ble_state, sl_si91x_performance_profile_t sl_si91x_wifi_state)
 {
-    return (wfx_rsi_power_save(sl_si91x_ble_state, sl_si91x_wifi_state) ? SL_STATUS_FAIL : SL_STATUS_OK);
+    int32_t error = rsi_bt_power_save_profile(sl_si91x_ble_state, 0);
```
lpbeliveau-silabs, line 868: Is there no way to merge these int32_t and sl_status_t like it was? Also, do we really need to handle negative values here?

lpbeliveau-silabs: Meaning `error` and `status`

src/platform/silabs/wifi/rs911x/WifiInterface.cpp
```diff
@@ -966,8 +933,26 @@ int32_t wfx_rsi_send_data(void * p, uint16_t len)
...
+sl_status_t wfx_power_save(void)
 {
-    return (wfx_rsi_power_save() ? SL_STATUS_FAIL : SL_STATUS_OK);
+    int32_t status;
```
lpbeliveau-silabs, line 938: Same as above, do we really need signedness?

lpbeliveau-silabs (approved):

src/platform/silabs/wifi/SiWx/WifiInterface.cpp
mkardous-silabs: `rsi_bt_power_save_profile` returns a negative value for error while sl_status_t codes are all positives so there is no clean way to merge them.

The previous impl used signed variable for both functions which i don't think is a good idea thus the split.
The only other option I see is to use `int32` for both and "hope" the sl_status_t value falls with the range which should be the case based on the size.

jmartinez-silabs (approved):

PR #112, opened 2024-11-16
[SL-ONLY] Move SiWx917 Sleep Calls to the SleepManager class

mkardous-silabs:
#### Description
The SleepManager class is now the responsible entity of managing the Wi-Fi sleep states based on the Matter SDK states.
PR moves the SiWx917 power saves class to the SleepManager. In this PR, the behavior isn't changed.

#### Tests
Manual tests with the 917 SoC that the different states still work
- Sleep after commissioning
- Sleep on power cycle

### DIFF
--- a/examples/platform/silabs/wifi/icd/SleepManager.cpp
+++ b/examples/platform/silabs/wifi/icd/SleepManager.cpp
@@ -1,4 +1,6 @@
 #include "SleepManager.h"
+#include <lib/support/logging/CHIPLogging.h>
+#include <platform/silabs/wifi/WifiInterfaceAbstraction.h>
 
 using namespace chip::app;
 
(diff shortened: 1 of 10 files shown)


examples/platform/silabs/wifi/icd/SleepManager.cpp
```diff
@@ -1,4 +1,6 @@
 #include "SleepManager.h"
+#include <lib/support/logging/CHIPLogging.h>
```
jmartinez-silabs, line 2: I wonder if the sleep manager should follow the icd states

mkardous-silabs: For the SIT ICD use-case, there isn't much benefit to follow the ICD states other than the basic Fast / slow poll behavior.
Where we need to be synced is with the LIT ICD behavior to optimize the Wi-Fi connectivity based on the ICD operationnal states.

jmartinez-silabs: Part of what i mean is that you are using a eventing based on commissioning events (open/close/completed) but really it is the same as following the kCommissioningWindowOpen KeepActiveFlagsValues. I suspect the wifi device will require to listen to not sleep and not filter mdns packets for any KeepActiveFlags

mkardous-silabs:
Closing PR to re-work the structure

mkardous-silabs:
Closing PR to re-work the structure

PR #113, opened 2024-11-18
[SL-UP] Introduce PlatformSleepManager to manage Wi-Fi sleep states

mkardous-silabs:
#### Description
PR introduces the PlatformSleepManager that is responsible of managing the Wi-Fi sleep states. The goal of this class is to centralize the buisness intelligence around sleep intelligence to avoid conflicting configuration.

PR doesn't changes the current behavior. It just moves the logic to a centralize location.

#### Tests
Manual tests


src/platform/silabs/wifi/SiWx/WifiInterface.cpp
```diff
@@ -338,14 +347,21 @@ sl_status_t SetWifiConfigurations()
     sl_status_t status = SL_STATUS_OK;
 
 #if CHIP_CONFIG_ENABLE_ICD_SERVER
-    // Setting the listen interval to 0 which will set it to DTIM interval
-    sl_wifi_listen_interval_t sleep_interval = { .listen_interval = 0 };
+    // [sl-only] Set the listen interval to the slow polling interval during association
+    sl_wifi_listen_interval_t sleep_interval = { .listen_interval =
+                                                     chip::ICDConfigurationData::GetInstance().GetSlowPollingInterval().count() };
```
chirag-silabs, line 352: I understand what you are trying to do here, but the slow polling interval count will not return 0 right which is needed to poll at DTIM 
This should be done during a subscription is started correct

chirag-silabs:
Hoping it is tested on wf200, since the platform sleep manager code is affecting RS9116 and SiWx917 soc/ncp

src/platform/silabs/wifi/SiWx/WifiInterface.cpp
mkardous-silabs: This doesn't exactly do that. To be able to change listen interval after associating with the ap, you need to specify to the AP your max listen interval for it to know how long to buffer message and how long before the AP disconnects you.

This is required for us to be able to change the listen interval at run time between 0 (DTIM) a and the Slow Poll configuration.

src/platform/silabs/wifi/icd/PlatformSleepManager.cpp
```diff
@@ -0,0 +1,126 @@
...
+namespace chip {
+namespace DeviceLayer {
+namespace Silabs {
+
+// Initialize the static instance
+PlatformSleepManager PlatformSleepManager::mInstance;
+
+CHIP_ERROR PlatformSleepManager::Init()
+{
+    return CHIP_NO_ERROR;
```
lpbeliveau-silabs, line 34: So I assume actions will eventually be implemented here, since we check that the init was successful in MatterConfig right?

src/platform/silabs/wifi/icd/PlatformSleepManager.cpp
```diff
@@ -0,0 +1,126 @@
...
+}
+
+void PlatformSleepManager::HandleCommissioningComplete()
+{
+    TransitionToLowPowerMode();
+}
+
+void PlatformSleepManager::HandleInternetConnectivityChange()
+{
+    if (!isCommissioningInProgress)
```
lpbeliveau-silabs, line 44: I do not have a clear state machine in mind, if we loose internet during commissioning, the path is that HandleCommissioningSessionStopped will be called and eventually and thus bring the chip back to sleep?

src/platform/silabs/wifi/icd/PlatformSleepManager.cpp
```diff
@@ -0,0 +1,126 @@
...
+}
+
+CHIP_ERROR PlatformSleepManager::RemoveHighPerformanceRequest()
+{
+    VerifyOrReturnError(mHighPerformanceRequestCounter > 0, CHIP_NO_ERROR,
+                        ChipLogError(DeviceLayer, "Wi-Fi configuration already in low power mode"));
+
+    mHighPerformanceRequestCounter--;
+
+    if (mHighPerformanceRequestCounter == 0)
```
lpbeliveau-silabs, line 88: This check is already inside `TransitionToLowPowerMode`

src/platform/silabs/wifi/icd/PlatformSleepManager.cpp
lpbeliveau-silabs, line 44: Since this check is already inside `TransitionToLowPowerMode` under certain condition, could we have a VerifyOrReturnValue in `TransitionToLowPowerMode` to just ignore the call there and reduce redundancy of checks?

src/platform/silabs/wifi/icd/PlatformSleepManager.cpp
```diff
@@ -0,0 +1,126 @@
...
+    {
+        ReturnErrorOnFailure(TransitionToLowPowerMode());
+    }
+
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR PlatformSleepManager::TransitionToLowPowerMode()
+{
+    VerifyOrReturnError(mHighPerformanceRequestCounter == 0, CHIP_NO_ERROR,
```
lpbeliveau-silabs, line 98: VerifyOrReturnError with `CHIP_NO_ERROR ` is misleading

lpbeliveau-silabs: How about `VerifyOrReturnValue`?

lpbeliveau-silabs (approved):

src/platform/silabs/wifi/icd/PlatformSleepManager.cpp
mkardous-silabs: This isn't a case that is currently supported by the existing wi-fi implementation so that would be an edge case that isn't covered.

The goal of the PR is just to move the existing logic to into a centralized spot.

src/platform/silabs/wifi/icd/PlatformSleepManager.cpp
mkardous-silabs: didn't about that define

src/platform/silabs/wifi/icd/PlatformSleepManager.cpp
mkardous-silabs: At some point yes - Once the Wi-Fi interface abstraction inti rework is done. The init should be moved there. For now, i don't which init does all patforms at the right time.

src/platform/silabs/wifi/icd/PlatformSleepManager.cpp
mkardous-silabs: Just moving the check into the TransitionToLowPowerMode "brakes" the go to sleep logic. I'll add a todo to re-work this for it to make sense.

As a whole, our go to sleep logic needs to be improved.

jmartinez-silabs (approved):

PR #122, opened 2024-11-21
[SL-ONLY] Silabs tracing integration

lpbeliveau-silabs:
Integrating our SillabsTracing with the matter tracing backend as well as with our bootup and ota.

Currently needs testing for OTA, timing at bootup is only registered as 0.

Matter backend only logs Sigma 2 and the end of case sessions, need to improve our instrumenting there.

In short, todos are:
- Integration in apps other than lighting app / refactor our base app to to it by default.
- Wapper around our System::SystemClock().GetMonotonicTimestamp() in SilabsTracer to use hardware timers before the Scheduler start
- Wrapper around our SilabsTracer calls to reduce all the #if calls
- Implementation of NVM storage for Watermarks and Maybe for certain traces
- Integrate with slc
- FIgure out metric_event_handler.slcc


src/tracing/silabs/BUILD.gn
```diff
@@ -0,0 +1,25 @@
...
+# limitations under the License.
+
+import("//build_overrides/build.gni")
+import("//build_overrides/chip.gni")
+
+config("tracing") {
+  include_dirs = [ "include" ]
+}
+
+source_set("silabs_tracing") {
```
jmartinez-silabs, line 22: I don't get this source_set

jmartinez-silabs (approved):

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -89,6 +89,12 @@
 #include "sl_cmp_config.h"
 #endif
 
+// Tracing
+#include <matter/tracing/build_config.h>
+#if MATTER_TRACING_ENABLED
```
mkardous-silabs, line 94:
```suggestion
#if SL_MATTER_TRACING_ENABLED
```
`SL_MATTER_` is the "standard" naming for slc, it will make it easier to integrate

examples/platform/silabs/BaseApplication.h
```diff
@@ -176,6 +177,12 @@ class BaseApplication
 protected:
     CHIP_ERROR Init();
 
+    /** @brief
+     * Function to be called at the end of Init to indicate that the application has completed its initialization.
+     * Currently only used for tracing, might want to move logging here as well in the future
+     */
+    void InitCompleteCallback(CHIP_ERROR err);
```
mkardous-silabs, line 184: I think it would be worth while to add a section for the param

examples/platform/silabs/MatterShell.cpp
```diff
@@ -112,6 +117,154 @@ void cmdSilabsInit()
 
 #endif // SL_CATALOG_CLI_PRESENT
 
+#if MATTER_TRACING_ENABLED
+using TimeTraceOperation = Tracing::Silabs::TimeTraceOperation;
+
+TimeTraceOperation StringToTimeTraceOperation(const char * str)
```
mkardous-silabs, line 123: Yes, the `StringToTimeTraceOperation` function can be simplified by using a `std::map` to map strings to `TimeTraceOperation` values. Here's how you can do it:

```cpp
#include <map>
#include <string>

// Define the map globally or inside the function
const std::map<std::string, TimeTraceOperation> timeTraceOperationMap = {
    {"Spake2p", TimeTraceOperation::kSpake2p},
    {"Pake1", TimeTraceOperation::kPake1},
    {"Pake2", TimeTraceOperation::kPake2},
    {"Pake3", TimeTraceOperation::kPake3},
    {"OperationalCredentials", TimeTraceOperation::kOperationalCredentials},
    {"AttestationVerification", TimeTraceOperation::kAttestationVerification},
    {"CSR", TimeTraceOperation::kCSR},
    {"NOC", TimeTraceOperation::kNOC},
    {"TransportLayer", TimeTraceOperation::kTransportLayer},
    {"TransportSetup", TimeTraceOperation::kTransportSetup},
    {"FindOperational", TimeTraceOperation::kFindOperational},
    {"CaseSession", TimeTraceOperation::kCaseSession},
    {"Sigma1", TimeTraceOperation::kSigma1},
    {"Sigma2", TimeTraceOperation::kSigma2},
    {"Sigma3", TimeTraceOperation::kSigma3},
    {"OTA", TimeTraceOperation::kOTA},
    {"ImageUpload", TimeTraceOperation::kImageUpload},
    {"ImageVerification", TimeTraceOperation::kImageVerification},
    {"AppApplyTime", TimeTraceOperation::kAppApplyTime},
    {"Bootup", TimeTraceOperation::kBootup},
    {"SilabsInit", TimeTraceOperation::kSilabsInit},
    {"MatterInit", TimeTraceOperation::kMatterInit},
    {"BufferFull", TimeTraceOperation::kBufferFull}
};

TimeTraceOperation StringToTimeTraceOperation(const char * str)
{
    auto it = timeTraceOperationMap.find(str);
    if (it != timeTraceOperationMap.end())
    {
        return it->second;
    }
    return TimeTraceOperation::kNumTraces;
}
```

This way, the function becomes more concise and easier to maintain.
Copilot: 1 - LP : 0

examples/platform/silabs/MatterShell.cpp
mkardous-silabs, line 1: Can we move all the tracing shell code to the shell/ directory with an TrancingShell.h/cpp?
This avoid having all the shells in a single files and makes the over structure cleaner.

examples/platform/silabs/MatterShell.cpp
```diff
@@ -112,6 +117,154 @@ void cmdSilabsInit()
...
+    {
+        return TimeTraceOperation::kBufferFull;
+    }
+    else
+    {
+        return TimeTraceOperation::kNumTraces;
+    }
+}
+
+CHIP_ERROR CmdTracingDispatch(int argc, char ** argv)
```
mkardous-silabs, line 223: You can have subcommands managed directly be the shell structures init. This avoid having huge if else statements to manages multiple commands.

examples/platform/silabs/BaseApplication.cpp
lpbeliveau-silabs: This is a CSA define.

src/tracing/silabs/BUILD.gn
lpbeliveau-silabs: it's the standard setup to map the maccro_imp file location to the matter tracing arg.gni

examples/platform/silabs/MatterShell.cpp
lpbeliveau-silabs: Except including map increases the code size and the ram usage, for the sole purpose of shrinking a few lines of code...

examples/platform/silabs/MatterShell.cpp
lpbeliveau-silabs: Maybe... but looking at src/lib/shell I don't see other companies putting their stuff in there. Also I prefer keeping silabs stuff into silabs folder.

examples/platform/silabs/MatterShell.cpp
lpbeliveau-silabs: Probably, I was just trying to imitate the style above.

Also:
> shell structures init.

Are you proposing we create a child class to the Shell class?

examples/platform/silabs/MatterShell.cpp
mkardous-silabs: Check examples/silabs/shell

examples/platform/silabs/MatterShell.cpp
jmartinez-silabs: He means that you can have SubCommands handler. That was done for LightSwitch and Door lock app shell commands.

Personally, I'm fine with either or. Check what Math means if you prefer it

src/platform/silabs/tracing/BackendImpl.h
```diff
@@ -0,0 +1,73 @@
...
+
+/** @brief
+ * Backend implementation for Silabs platform acting as an interface with SilabsTracer and the matter layer.
+ */
+class BackendImpl : public ::chip::Tracing::Backend
+{
+public:
+    BackendImpl() = default;
+    // TraceBegin, TraceEnd and TraceInstant are redundant with LogMetricEvent in the usecases that we are trying to track,
+    // so we are not implementing them at current to avoid duplication of the same information in the trace. We might implement it
```
mkardous-silabs, line 49:
```suggestion
    // so we are not implementing these at the moment to avoid duplication of the same information in the trace. We might implement them
```

src/platform/silabs/tracing/include/matter/tracing/macros_impl.h
```diff
@@ -0,0 +1,23 @@
...
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+#pragma once
+
+/* Ensure we do not have double tracing macros defined */
+#if defined(MATTER_TRACE_BEGIN)
+#error "Tracing macros seem to be double defined"
```
mkardous-silabs, line 22:
```suggestion
#error "Tracing macro seems to be double defined"
```

src/tracing/silabs/include/matter/tracing/macros_impl.h
```diff
@@ -0,0 +1,25 @@
...
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+#pragma once
+
+/* Ensure we do not have double tracing macros defined */
+#if defined(MATTER_TRACE_BEGIN)
+#error "Tracing macros seem to be double defined"
```
mkardous-silabs, line 22:
```suggestion
#error "Tracing macro seems to be double defined"
```

examples/platform/silabs/shell/tracing/TracingShellCommands.cpp
```diff
@@ -0,0 +1,203 @@
...
+CHIP_ERROR FlushCommandHandler(int argc, char ** argv)
+{
+    CHIP_ERROR error = CHIP_NO_ERROR;
+    if (strcmp(argv[0], "all") == 0)
+    {
+        error = Tracing::Silabs::SilabsTracer::Instance().TraceBufferFlushAll();
+    }
+    else
+    {
+        TimeTraceOperation operation = StringToTimeTraceOperation(argv[1]);
```
mkardous-silabs, line 179: What happens if the operation doesn't exist in the string to time function?

In practice, it would be interesting to have a second layer of commands that defines all the operations so that they can be printed with a help menu.

lpbeliveau-silabs: For the watermarks, it will output an error, for the flush, it won't output anything since it won't find a trace for the kNumTrace event, I'll implement a LisTimeTraceOperations method to display this.

src/platform/silabs/tracing/SilabsTracing.cpp
```diff
@@ -286,7 +286,7 @@ CHIP_ERROR SilabsTracer::OutputTrace(const TimeTracker & tracker)
 CHIP_ERROR SilabsTracer::OutputWaterMark(TimeTraceOperation aOperation)
 {
     VerifyOrReturnError(to_underlying(aOperation) < kNumTraces, CHIP_ERROR_INVALID_ARGUMENT,
-                        ChipLogError(DeviceLayer, "Invalid watermark"));
+                        ChipLogError(DeviceLayer, "Invalid TimeTraceOperation"));
```
mkardous-silabs, line 289: just maing sure this is an ok change since due to the function name

mkardous-silabs (approved):

src/platform/silabs/tracing/SilabsTracing.cpp
lpbeliveau-silabs: Yes, watermarks and traces are both mapped using TimeTraceOperation

examples/platform/silabs/efr32/BUILD.gn
```diff
@@ -141,7 +145,10 @@ config("efr32-common-config") {
 }
 
 source_set("efr32-common") {
-  deps = [ "${silabs_common_plat_dir}/provision:storage" ]
+  deps = [
+    "${chip_root}/src/tracing",
```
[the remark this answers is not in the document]
lpbeliveau-silabs: Short answer yes, because of the define, but I'll look around to check if we could have a cleaner dependency that doesn't involve the registry source.

src/platform/silabs/tracing/Diagram.mmd
```diff
@@ -0,0 +1,65 @@
...
+        +SingleLinkedListNode* mpNext
+    }
+
+    SilabsTracer --> TimeTracker
+    SilabsTracer --> Watermark
+    SilabsTracer --> TimeTrackerList
+    TimeTrackerList --> SingleLinkedListNode
+    SingleLinkedListNode --> TimeTracker
+```mermaid
+````
```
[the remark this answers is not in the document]
lpbeliveau-silabs: We don't need the diagrams

src/platform/silabs/tracing/SequenceDiagram.mmd
```diff
@@ -0,0 +1,16 @@
...
+    participant Watermark
+
+    Client->>SilabsTracer: TimeTraceInstant(aOperation, error)
+    SilabsTracer->>TimeTracker: Create tracker
+    SilabsTracer->>SilabsTracer: OutputTrace(tracker)
+    SilabsTracer->>SilabsTracer: OutputTimeTracker(tracker)
+    SilabsTracer->>Watermark: Update Watermark
+    SilabsTracer->>Client: Return result
+
+```mermaid
```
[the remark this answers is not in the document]
lpbeliveau-silabs: same here, no we don't, but we don't need the diagrams at all.

examples/platform/silabs/efr32/BUILD.gn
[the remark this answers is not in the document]
lpbeliveau-silabs: Yup, not needed

PR #123, opened 2024-11-22
[CSA-CP] Adds changes for min and max values of refrigerator temperature controlled cabinet (#36592)

shgutte:
https://github.com/project-chip/connectedhomeip/pull/36592


mkardous-silabs (approved):

examples/refrigerator-app/silabs/src/RefrigeratorManager.cpp
```diff
@@ -102,7 +102,6 @@ void RefrigeratorManager::TempCtrlAttributeChangeHandler(EndpointId endpointId,
     case TempCtrlAttr::TemperatureSetpoint::Id: {
         int16_t temperatureSetpoint = ConvertToPrintableTemp(static_cast<int16_t>(*value));
         mTemperatureSetpoint        = temperatureSetpoint;
-        TempCtrlAttr::TemperatureSetpoint::Set(endpointId, temperatureSetpoint);
```
jmartinez-silabs, line 105: is this set at some point somewhere else now?

shgutte: The function `emberAfTemperatureControlClusterSetTemperatureCallback` in `temperature-control-server.cpp` is called internally and does not need to be invoked from the application layer. Calling it directly from the application layer creates a loop, which results in incorrect updates to the temperature setpoint. Therefore, it is being removed.

jmartinez-silabs (approved):

PR #126, opened 2024-11-22
[SL-UP] Remove WF200 define duplication to avoid value conflicts at runtime

mkardous-silabs:
#### Description
When additional Wi-Fi platforms were added, WF200 defines and constructs were used for "non-silabs" platforms. During the Wi-Fi abstraction work, an effort was done to reduce the number of location where defines we set. 

Since the names were re-used with defines and the Si SDK defines thses names as typedef enums, there were build issue to flag that there could be values mismatchs.

PR moves the defines in a gatted section that is not used by the WF200 to avoid anything conflicting values.

#### Tests
Manuel tests to validates that WF200 can join a Wi-Fi network

### DIFF
--- a/src/platform/silabs/wifi/wfx_msgs.h
+++ b/src/platform/silabs/wifi/wfx_msgs.h
@@ -26,6 +26,15 @@
 #include "sl_wfx_api.h"
 #include "sl_wfx_constants.h"
 #else
+
+// These names exists in the Si SDK as typedef enum. If they are present in the WF200 builds, we end up with conflicting
+// definitions but no erros because one is a define the other is a typedef enum. This causes different files to use different
+// values.
+#define SL_WFX_STARTUP_IND_ID (1)
+#define SL_WFX_CONNECT_IND_ID (2)
+#define SL_WFX_DISCONNECT_IND_ID (3)
+#define SL_WFX_SCAN_COMPLETE_ID (4)
+
 typedef struct
 {
     uint8_t octet[6]; ///< Table to store a MAC address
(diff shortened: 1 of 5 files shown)


src/platform/silabs/wifi/wfx_msgs.h
```diff
@@ -26,6 +26,15 @@
 #include "sl_wfx_api.h"
 #include "sl_wfx_constants.h"
 #else
+
+// These names exists in the Si SDK as typedef enum. If they are present in the WF200 builds, we end up with conflicting
+// definitions but no erros because one is a define the other is a typedef enum. This causes different files to use different
+// values.
+#define SL_WFX_STARTUP_IND_ID (1)
```
jmartinez-silabs, line 33: we should rename them at some point

jmartinez-silabs (approved):

third_party/silabs/efr32_sdk.gni
```diff
@@ -1022,7 +1010,8 @@ template("efr32_sdk") {
       sources += [
         "${efr32_sdk_root}/platform/radio/wifi/wfx_fmac_driver/bus/sl_wfx_bus.c",
         "${efr32_sdk_root}/platform/radio/wifi/wfx_fmac_driver/bus/sl_wfx_bus_spi.c",
-        "${efr32_sdk_root}/platform/radio/wifi/wfx_fmac_driver/secure_link/sl_wfx_secure_link.c",
+
+        # "${efr32_sdk_root}/platform/radio/wifi/wfx_fmac_driver/secure_link/sl_wfx_secure_link.c",
```
mykrupp, line 1014: Can the commented code be removed? Or have a TODO to remove/uncomment it

mykrupp (approved):

src/platform/silabs/wifi/wfx_msgs.h
mkardous-silabs: Agreed; we need to have a matter abstraction that doesn't re-use all the wf200 structures.

PR #128, opened 2024-11-25
[SL-UP] Fix builds for sleep manager

bhmanda-silabs:
- While adding wifi sleep manager component in extension , build and copy contents are failing for EFR and SoC platforms.
- Added CHIP_CONFIG_ENABLE_ICD_SERVER macro, wherever WifiSleepManager.h is included.

### DIFF
--- a/src/platform/silabs/wifi/SiWx/WifiInterface.cpp
+++ b/src/platform/silabs/wifi/SiWx/WifiInterface.cpp
@@ -33,7 +33,6 @@
 #include "sl_status.h"
 #include "sl_wifi_device.h"
 #include "task.h"
-#include <app/icd/server/ICDConfigurationData.h>
 #include <app/icd/server/ICDServerConfig.h>
 #include <inet/IPAddress.h>
 #include <lib/support/CHIPMem.h>
@@ -72,6 +71,7 @@ extern "C" {
 #endif
 
 #if CHIP_CONFIG_ENABLE_ICD_SERVER
+#include <app/icd/server/ICDConfigurationData.h>
 #include <platform/silabs/wifi/icd/WifiSleepManager.h>
 
 #if SLI_SI91X_MCU_INTERFACE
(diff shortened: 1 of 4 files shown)


mkardous-silabs (approved):

src/platform/silabs/wifi/SiWx/WifiInterface.cpp
```diff
@@ -72,6 +71,7 @@ extern "C" {
 #endif
 
 #if CHIP_CONFIG_ENABLE_ICD_SERVER
+#include <app/icd/server/ICDConfigurationData.h>
```
lpbeliveau-silabs, line 74: So it seems we are moving everything ICD related under this `#if CHIP_CONFIG_ENABLE_ICD_SERVER`, why not also move `#include <app/icd/server/ICDServerConfig.h>`?

lpbeliveau-silabs (approved):
Approved but in the long term we might want to look into implementing WifiSleepManager in a way that is is stubbed when not building for ICD rather than relying on having #if conditions everywhere.

src/platform/silabs/wifi/SiWx/WifiInterface.cpp
mkardous-silabs: ICDServerConfig.h is the header that inlucdes the `CHIP_CONFIG_ENABLE_ICD_SERVER` define so it always needs to be included before using the define wether its an icd build or not.

bhmanda-silabs: Yeah, many files has this header file included but not under this macro like MatterConfig.cpp,demo-ui.cpp, so better take it as separate task instead of adding in the same PR.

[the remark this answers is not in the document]
bhmanda-silabs: Added //nogncheck @mkardous-silabs

PR #137, opened 2024-11-27
[SL-UP] Adds and updates logs for RS911X

rosahay-silabs:
This PR aims to add error status codes for RS911x with respect to MATTER-4193


mkardous-silabs (approved):
If are updating the error / progress level of certain logs, can we double check the other prints to make sure they have the correct level as well?

src/platform/silabs/wifi/rs911x/WifiInterface.cpp
mkardous-silabs, line 1: What is the change in this file? This doesn't seem related to what the initial goal of the PR was?

mkardous-silabs: ok thanks for the information! I was surprise to see the VerifyOrReturn being added - seemed like a bug fix

mkardous-silabs (approved):

chirag-silabs (approved):

senthilku (approved):

PR #139, opened 2024-11-27
[SL-ONLY] Add business logic for the selective listening impl

mkardous-silabs:
#### Description
PR implements the `CanGoToLIBasedSleep` which is the core of the of the Selective Listening Design.
The initial logic is to check that there is at least 1 subscription per fabric and that the commissioning window is not opened to enable LI based sleep.

PR removes the dependency on the CommissioningWindowManager because the window is considered open while the 
`OnCommissioningWindowClosed` callback is being processed.

#### Tests
Manual tests to validate that device is in LI / DTIm based sleep at the right moments. The way the Wi-Fi integration is implemented prevents from having unit tests. This will be in a follow up once the platform layer is improved.

### DIFF
--- a/examples/platform/silabs/wifi/icd/ApplicationSleepManager.cpp
+++ b/examples/platform/silabs/wifi/icd/ApplicationSleepManager.cpp
@@ -29,8 +29,6 @@ CHIP_ERROR ApplicationSleepManager::Init()
     VerifyOrReturnError(mFabricTable != nullptr, CHIP_ERROR_INVALID_ARGUMENT, ChipLogError(AppServer, "FabricTable is null"));
     VerifyOrReturnError(mSubscriptionsInfoProvider != nullptr, CHIP_ERROR_INVALID_ARGUMENT,
                         ChipLogError(AppServer, "SubscriptionsInfoProvider is null"));
-    VerifyOrReturnError(mCommissioningWindowManager != nullptr, CHIP_ERROR_INVALID_ARGUMENT,
-                        ChipLogError(AppServer, "CommissioningWindowManager is null"));
     VerifyOrReturnError(mWifiSleepManager != nullptr, CHIP_ERROR_INVALID_ARGUMENT,
                         ChipLogError(AppServer, "WifiSleepManager is null"));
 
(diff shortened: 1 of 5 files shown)


examples/platform/silabs/wifi/icd/ApplicationSleepManager.h
```diff
@@ -94,10 +102,14 @@ class ApplicationSleepManager : public chip::app::ReadHandler::ApplicationCallba
...
+     * @brief Function encapsulates the application logic to determine if the Wi-Fi device can go into LI based sleep.
      *
-     * @return true
-     * @return false
+     *        - 1. Check if the commissioning window is open. If it is open, the Wi-Fi device cannot go to LI based sleep.
+     *        - 2. Check if all Fabrics have at least 1 subscription. If there is at least one fabric without a subscription, the
+     *             Wi-Fi cannot go to LI based sleep.
+     *
+     * @return true if the device can go to LI sleep
+     * @return false if the device cannot go to LI sleep
```
lpbeliveau-silabs, line 112:
```suggestion
     * @return true if the device can go to LI sleep
     *         false if the device cannot go to LI sleep
```

lpbeliveau-silabs: Copilot seems to like these double `@return` lines ...

lpbeliveau-silabs (approved):

examples/platform/silabs/wifi/icd/ApplicationSleepManager.cpp
```diff
@@ -71,10 +81,28 @@ void ApplicationSleepManager::OnFabricCommitted(const chip::FabricTable & fabric
...
+    bool canGoToLIBasedSleep = true;
+
+    if (mIsCommissionningWindowOpen)
+    {
+        ChipLogProgress(AppServer, "Commissioning Window is Open - Cannot go to LI based sleep");
+        canGoToLIBasedSleep = false;
+    }
+    else
+    {
+        for (ConstFabricIterator it = mFabricTable->begin(); it != mFabricTable->end(); ++it)
```
lpbeliveau-silabs, line 93:
```suggestion
        for (auto it = mFabricTable->begin(); it != mFabricTable->end(); ++it)
```

chirag-silabs:
Just a suggestion, since we are adding the sleep requirement on window open and window close which can be changed, since this is only needed during the commissioning. After commissioning if window is opened or closed the device can be in the DTIM sleep.

mkardous-silabs:
> Just a suggestion, since we are adding the sleep requirement on window open and window close which can be changed, since this is only needed during the commissioning. After commissioning if window is opened or closed the device can be in the DTIM sleep.

@chirag-silabs I don't follow your suggestion

rcasallas-silabs (approved):

PR #142, opened 2024-11-27
[SL-UP] Bugfix/silabs out rtt removal

lpbeliveau-silabs:
`SILABS_LOG_OUT_RTT ` 
Had not been added to any slcc files, it was therefore always set to true in Logging.cpp.

Since we are already using SILABS_LOG_OUT_UART as the conditionto toggle between both, I chose to completely remove 
SILABS_LOG_OUT_RTT as it was redundant. 

We could just as well update our logging components to add a define to SILABS_LOG_OUT_RTT's value in there though.

### DIFF
--- a/third_party/silabs/SiWx917_sdk.gni
+++ b/third_party/silabs/SiWx917_sdk.gni
@@ -334,10 +334,9 @@ template("siwx917_sdk") {
     }
 
     if (sl_uart_log_output) {
-      defines += [
-        "SILABS_LOG_OUT_UART=1",
-        "SILABS_LOG_OUT_RTT=0",
-      ]
+      defines += [ "SILABS_LOG_OUT_UART=1" ]
+    } else {
+      defines += [ "SILABS_LOG_OUT_UART=0" ]
     }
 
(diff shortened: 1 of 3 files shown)


mkardous-silabs (approved):

third_party/silabs/SiWx917_sdk.gni
```diff
@@ -334,10 +334,9 @@ template("siwx917_sdk") {
     }
 
     if (sl_uart_log_output) {
```
rcasallas-silabs, line 336: Maybe:
`defines += [ "SILABS_LOG_OUT_UART=${sl_uart_log_output}" ]`

third_party/silabs/efr32_sdk.gni
```diff
@@ -490,10 +490,9 @@ template("efr32_sdk") {
     }
 
     if (sl_uart_log_output) {
```
rcasallas-silabs, line 492: Maybe:
`defines += [ "SILABS_LOG_OUT_UART=${sl_uart_log_output}" ]`

rcasallas-silabs (approved):

PR #148, opened 2024-11-29
[SL-UP]Adding the dimmer switch App

chirag-silabs:
This PR adds the dimmer-switch application by extending the existing light-switch application.

The zap file of light-switch app is modified by adding level-control-cluster client.The levelcontrol cluster can be controlled by sending commands from shell.

Button switch level control will be added in a follow up PR

**Tested:**
Tested locally with a lighting app device levelcontrol commands handled by the switch

### DIFF
--- a/examples/light-switch-app/silabs/include/AppEvent.h
+++ b/examples/light-switch-app/silabs/include/AppEvent.h
@@ -30,6 +30,13 @@ struct AppEvent
         kEventType_Timer,
         kEventType_Light,
         kEventType_Install,
+        kEventType_ResetWarning,
+        kEventType_ResetCanceled,
+        // Button events
+        kEventType_UpPressed,
+        kEventType_UpReleased,
+        kEventType_DownPressed,
+        kEventType_DownReleased,
     };
 
     uint16_t Type;
@@ -49,6 +56,10 @@ struct AppEvent
             uint8_t Action;
             int32_t Actor;
         } LightEvent;
+        struct
+        {
+            void * Context;
+        } LightSwitchEvent;
     };
 
     EventHandler Handler;
--- a/examples/light-switch-app/silabs/include/AppEvent.h
+++ b/examples/light-switch-app/silabs/include/AppEvent.h
@@ -53,11 +53,6 @@ struct AppEvent
         } TimerEvent;
         struct
         {
-            uint8_t Action;
-            int32_t Actor;
-        } LightEvent;
-        struct
-        {
             void * Context;
         } LightSwitchEvent;
     };
--- a/examples/light-switch-app/silabs/include/BindingHandler.h
+++ b/examples/light-switch-app/silabs/include/BindingHandler.h
@@ -19,16 +19,57 @@
 #include "app-common/zap-generated/ids/Clusters.h"
 #include "app-common/zap-generated/ids/Commands.h"
 #include "lib/core/CHIPError.h"
+#include <platform/CHIPDeviceLayer.h>
+#include <app/clusters/bindings/bindings.h>
+#include <variant>
+
+using namespace chip;
+using namespace chip::app;
+using namespace chip::app::Clusters::LevelControl;
 
 CHIP_ERROR InitBindingHandler();
 void SwitchWorkerFunction(intptr_t context);
 void BindingWorkerFunction(intptr_t context);
+
 
 struct BindingCommandData
 {
     chip::EndpointId localEndpointId = 1;
     chip::CommandId commandId;
     chip::ClusterId clusterId;
-    bool isGroup         = false;
-    uint32_t args[7];
+    bool isGroup = false;
+
+    struct MoveToLevel
+    {
+        uint8_t level;
+        DataModel::Nullable<uint16_t> transitionTime;
+        chip::BitMask<OptionsBitmap> optionsMask;
+        chip::BitMask<OptionsBitmap> optionsOverride;
+    };
+
+    struct Move
+    {
+        MoveModeEnum moveMode;
+        DataModel::Nullable<uint8_t> rate;
+        chip::BitMask<OptionsBitmap> optionsMask;
+        chip::BitMask<OptionsBitmap> optionsOverride;
+    };
+
+    struct Step
+    {
+        StepModeEnum stepMode;
+        uint8_t stepSize;
+        DataModel::Nullable<uint16_t> transitionTime;
+        chip::BitMask<OptionsBitmap> optionsMask;
+        chip::BitMask<OptionsBitmap> optionsOverride;
+    };
+
+    struct Stop
+    {
+        chip::BitMask<OptionsBitmap> optionsMask;
+        chip::BitMask<OptionsBitmap> optionsOverride;
+    };
+
+    // Use std::variant to replace the union
+    std::variant<MoveToLevel, Move, Step, Stop> commandData;
 };
--- a/examples/light-switch-app/silabs/include/BindingHandler.h
+++ b/examples/light-switch-app/silabs/include/BindingHandler.h
@@ -32,6 +32,16 @@ void SwitchWorkerFunction(intptr_t context);
 void BindingWorkerFunction(intptr_t context);
 
 
+struct CommandBase
+{
+    chip::BitMask<OptionsBitmap> optionsMask;
+    chip::BitMask<OptionsBitmap> optionsOverride;
+
+    // Constructor to initialize the BitMask
+    CommandBase()
+        : optionsMask(0), optionsOverride(0) {}
+};
+
 struct BindingCommandData
 {
     chip::EndpointId localEndpointId = 1;
@@ -39,37 +49,26 @@ struct BindingCommandData
     chip::ClusterId clusterId;
     bool isGroup = false;
 
-    struct MoveToLevel
+    struct MoveToLevel : public CommandBase
     {
         uint8_t level;
         DataModel::Nullable<uint16_t> transitionTime;
-        chip::BitMask<OptionsBitmap> optionsMask;
-        chip::BitMask<OptionsBitmap> optionsOverride;
     };
-
-    struct Move
+    struct Move : public CommandBase
     {
         MoveModeEnum moveMode;
         DataModel::Nullable<uint8_t> rate;
-        chip::BitMask<OptionsBitmap> optionsMask;
-        chip::BitMask<OptionsBitmap> optionsOverride;
     };
-
-    struct Step
+    struct Step : public CommandBase
     {
         StepModeEnum stepMode;
         uint8_t stepSize;
         DataModel::Nullable<uint16_t> transitionTime;
-        chip::BitMask<OptionsBitmap> optionsMask;
-        chip::BitMask<OptionsBitmap> optionsOverride;
     };
-
-    struct Stop
+    struct Stop : public CommandBase
     {
-        chip::BitMask<OptionsBitmap> optionsMask;
-        chip::BitMask<OptionsBitmap> optionsOverride;
+        // Inherits optionsMask and optionsOverride from CommandBase
     };
-
-    // Use std::variant to replace the union
+    // Use std::variant to hold different command types
     std::variant<MoveToLevel, Move, Step, Stop> commandData;
 };
--- a/examples/light-switch-app/silabs/include/BindingHandler.h
+++ b/examples/light-switch-app/silabs/include/BindingHandler.h
@@ -29,5 +29,6 @@ struct BindingCommandData
     chip::EndpointId localEndpointId = 1;
     chip::CommandId commandId;
     chip::ClusterId clusterId;
-    bool isGroup = false;
+    bool isGroup         = false;
+    uint32_t args[7];
 };
--- a/examples/light-switch-app/silabs/include/LightSwitchMgr.h
+++ b/examples/light-switch-app/silabs/include/LightSwitchMgr.h
@@ -23,6 +23,10 @@
 #include <lib/core/CHIPError.h>
 #include <lib/support/CodeUtils.h>
 
+#include "AppEvent.h"
+#include <cmsis_os2.h>
+#include <string>
+
 class LightSwitchMgr
 {
 public:
@@ -38,21 +42,63 @@ public:
         chip::EndpointId endpoint;
         chip::EventId event;
     };
+    struct Timer
+    {
+        typedef void (*Callback)(Timer & timer);
 
+        Timer(uint32_t timeoutInMs, Callback callback, void * context);
+        ~Timer();
+
+        void Start();
+        void Stop();
+        void Timeout();
+
+        Callback mCallback = nullptr;
+        void * mContext    = nullptr;
+        bool mIsActive     = false;
+
+        osTimerId_t mHandler = nullptr;
+
+    private:
+        static void TimerCallback(void * timerCbArg);
+    };
     CHIP_ERROR Init(chip::EndpointId lightSwitchEndpoint, chip::EndpointId genericSwitchEndpoint);
 
     void GenericSwitchOnInitialPress();
     void GenericSwitchOnShortRelease();
 
     void TriggerLightSwitchAction(LightSwitchAction action, bool isGroupCommand = false);
+    void TriggerLevelControlAction(uint8_t stepMode, bool isGroupCommand = false);
 
     static LightSwitchMgr & GetInstance() { return sSwitch; }
 
+    LightSwitchMgr();
+
+    static void ButtonEventHandler(uint8_t button, uint8_t btnAction);
+
+    static void GeneralEventHandler(AppEvent * aEvent);
+
+protected:
+
+    static void OnLongPressTimeout(Timer & timer);
+
+    Timer * mLongPressTimer = nullptr;
+    bool mDownPressed       = false;
+    bool mResetWarning      = false;
+
 private:
     static LightSwitchMgr sSwitch;
+    void HandleLongPress();
 
     static void GenericSwitchWorkerFunction(intptr_t context);
 
     chip::EndpointId mLightSwitchEndpoint   = chip::kInvalidEndpointId;
     chip::EndpointId mGenericSwitchEndpoint = chip::kInvalidEndpointId;
+    /**
+     * @brief Button event processing function
+     *        Function triggers a switch action sent to the CHIP task
+     *
+     * @param aEvent button event being processed
+     */
+    static void SwitchActionEventHandler(AppEvent * aEvent);
 };
--- a/examples/light-switch-app/silabs/include/LightSwitchMgr.h
+++ b/examples/light-switch-app/silabs/include/LightSwitchMgr.h
@@ -22,11 +22,17 @@
 #include <app/util/basic-types.h>
 #include <lib/core/CHIPError.h>
 #include <lib/support/CodeUtils.h>
+#include <app/clusters/bindings/bindings.h>
+#include <platform/CHIPDeviceLayer.h>
+#include "app-common/zap-generated/ids/Clusters.h"
 
 #include "AppEvent.h"
 #include <cmsis_os2.h>
 #include <string>
 
+using namespace chip;
+using namespace chip::app;
+using namespace chip::app::Clusters::LevelControl;
 class LightSwitchMgr
 {
 public:
@@ -42,6 +48,12 @@ public:
         chip::EndpointId endpoint;
         chip::EventId event;
     };
+
+    static constexpr uint8_t stepSize = 1;
+    static constexpr DataModel::Nullable<uint16_t> transitionTime = 0;
+    static constexpr chip::BitMask<OptionsBitmap> optionsMask = 0;
+    static constexpr chip::BitMask<OptionsBitmap> optionsOverride = 0;
+
     struct Timer
     {
         typedef void (*Callback)(Timer & timer);
@@ -62,6 +74,9 @@ public:
     private:
         static void TimerCallback(void * timerCbArg);
     };
+
+    LightSwitchMgr() = default;
+
     CHIP_ERROR Init(chip::EndpointId lightSwitchEndpoint, chip::EndpointId genericSwitchEndpoint);
 
     void GenericSwitchOnInitialPress();
@@ -71,8 +86,6 @@ public:
     void TriggerLevelControlAction(uint8_t stepMode, bool isGroupCommand = false);
 
     static LightSwitchMgr & GetInstance() { return sSwitch; }
-
-    LightSwitchMgr();
 
     static void ButtonEventHandler(uint8_t button, uint8_t btnAction);
 
--- a/examples/light-switch-app/silabs/include/LightSwitchMgr.h
+++ b/examples/light-switch-app/silabs/include/LightSwitchMgr.h
@@ -48,10 +48,12 @@ public:
         chip::EventId event;
     };
 
-    static constexpr uint8_t stepSize = 1;
-    static constexpr DataModel::Nullable<uint16_t> transitionTime = 0;
-    static constexpr chip::BitMask<OptionsBitmap> optionsMask = 0;
-    static constexpr chip::BitMask<OptionsBitmap> optionsOverride = 0;
+    static constexpr Clusters::LevelControl::Commands::Step::Type stepCommand = {
+        .stepSize = 1,
+        .transitionTime = 0,
+        .optionsMask = 0,
+        .optionsOverride = 0
+    };
 
     struct Timer
     {
@@ -74,32 +76,35 @@ public:
         static void TimerCallback(void * timerCbArg);
     };
 
-    LightSwitchMgr() = default;
-
     CHIP_ERROR Init(chip::EndpointId lightSwitchEndpoint, chip::EndpointId genericSwitchEndpoint);
 
     void GenericSwitchOnInitialPress();
     void GenericSwitchOnShortRelease();
 
     void TriggerLightSwitchAction(LightSwitchAction action, bool isGroupCommand = false);
-    void TriggerLevelControlAction(uint8_t stepMode, bool isGroupCommand = false);
+    void TriggerLevelControlAction(StepModeEnum stepMode, bool isGroupCommand = false);
+
+    AppEvent CreateNewEvent(AppEvent::AppEventTypes type);
 
     static LightSwitchMgr & GetInstance() { return sSwitch; }
 
     static void ButtonEventHandler(uint8_t button, uint8_t btnAction);
 
-    static void GeneralEventHandler(AppEvent * aEvent);
+    static void AppEventHandler(AppEvent * aEvent);
 
-protected:
+// protected:
 
-    static void OnLongPressTimeout(Timer & timer);
+
+private:
+    static LightSwitchMgr sSwitch;
 
     Timer * mLongPressTimer = nullptr;
     bool mDownPressed       = false;
     bool mResetWarning      = false;
 
-private:
-    static LightSwitchMgr sSwitch;
+    static void OnLongPressTimeout(Timer & timer);
+    LightSwitchMgr() = default;
+
     void HandleLongPress();
 
     static void GenericSwitchWorkerFunction(intptr_t context);
--- a/examples/light-switch-app/silabs/include/LightSwitchMgr.h
+++ b/examples/light-switch-app/silabs/include/LightSwitchMgr.h
@@ -24,8 +24,7 @@
 #include <lib/support/CodeUtils.h>
 #include <app/clusters/bindings/bindings.h>
 #include <platform/CHIPDeviceLayer.h>
-#include "app-common/zap-generated/ids/Clusters.h"
-
+#include <app-common/zap-generated/ids/Clusters.h>
 #include "AppEvent.h"
 #include <cmsis_os2.h>
 #include <string>
@@ -107,7 +106,8 @@ private:
 
     chip::EndpointId mLightSwitchEndpoint   = chip::kInvalidEndpointId;
     chip::EndpointId mGenericSwitchEndpoint = chip::kInvalidEndpointId;
-    /**
+   
+     /**
      * @brief Button event processing function
      *        Function triggers a switch action sent to the CHIP task
      *
--- a/examples/light-switch-app/silabs/include/LightSwitchMgr.h
+++ b/examples/light-switch-app/silabs/include/LightSwitchMgr.h
@@ -88,12 +88,17 @@ public:
 
     static LightSwitchMgr & GetInstance() { return sSwitch; }
 
+    /**
+     * @brief Event handler when a button is pressed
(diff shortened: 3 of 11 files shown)


examples/light-switch-app/silabs/src/BindingHandler.cpp
```diff
@@ -85,6 +88,181 @@ void ProcessOnOffGroupBindingCommand(CommandId commandId, const EmberBindingTabl
...
+    VerifyOrDie(peer_device != nullptr && peer_device->ConnectionReady());
+
+    Clusters::LevelControl::Commands::MoveToLevel::Type moveToLevelCommand;
+    Clusters::LevelControl::Commands::Move::Type moveCommand;
+    Clusters::LevelControl::Commands::Step::Type stepCommand;
+    Clusters::LevelControl::Commands::Stop::Type stopCommand;
+    Clusters::LevelControl::Commands::MoveToLevelWithOnOff::Type moveToLevelWithOnOffCommand;
+    Clusters::LevelControl::Commands::MoveWithOnOff::Type moveWithOnOffCommand;
+    Clusters::LevelControl::Commands::StepWithOnOff::Type stepWithOnOffCommand;
+    Clusters::LevelControl::Commands::StopWithOnOff::Type stopWithOnOffCommand;
```
jmartinez-silabs, line 111: why do we create all those variable before their use? Why not set it on the actual case so they are not instantiate when it is not needed

examples/light-switch-app/silabs/src/BindingHandler.cpp
```diff
@@ -85,6 +88,181 @@ void ProcessOnOffGroupBindingCommand(CommandId commandId, const EmberBindingTabl
...
+                                         stopWithOnOffCommand, onSuccess, onFailure);
+        break;
+    }
+}
+
+void ProcessLevelControlGroupBindingCommand(BindingCommandData * data, const EmberBindingTableEntry & binding)
+{
+    Messaging::ExchangeManager & exchangeMgr = Server::GetInstance().GetExchangeManager();
+
+    Clusters::LevelControl::Commands::MoveToLevel::Type moveToLevelCommand;
```
jmartinez-silabs, line 191: same comment here

examples/light-switch-app/silabs/src/ShellCommands.cpp
```diff
@@ -238,6 +240,542 @@ CHIP_ERROR GroupToggleSwitchCommandHandler(int argc, char ** argv)
...
+    data->commandId           = Clusters::LevelControl::Commands::StopWithOnOff::Id;
+    data->clusterId           = Clusters::LevelControl::Id;
+    data->args[0]             = atoi(argv[0]);
+    data->args[1]             = atoi(argv[1]);
+
+    DeviceLayer::PlatformMgr().ScheduleWork(SwitchWorkerFunction, reinterpret_cast<intptr_t>(data));
+    return CHIP_NO_ERROR;
+}
+
+#if 0
```
jmartinez-silabs, line 413: I don't think we need them other than maybe read CurrentLevel

examples/light-switch-app/silabs/src/ShellCommands.cpp
```diff
@@ -238,6 +240,542 @@ CHIP_ERROR GroupToggleSwitchCommandHandler(int argc, char ** argv)
...
+{
+    if (argc != 4)
+    {
+        return LevelControlHelpHandler(argc, argv);
+    }
+
+    BindingCommandData * data = Platform::New<BindingCommandData>();
+    data->commandId           = Clusters::LevelControl::Commands::MoveToLevel::Id;
+    data->clusterId           = Clusters::LevelControl::Id;
+    data->args[0]             = atoi(argv[0]);
```
jmartinez-silabs, line 273: `atoi()` documentation says it is deprecated in favour of `strtol()`

examples/light-switch-app/silabs/src/ShellCommands.cpp
arun-silabs: BindingHandler is not implemented for read commands, should it be done for currentLevel?

examples/light-switch-app/silabs/src/BindingHandler.cpp
arun-silabs: moved inside the actual cases

examples/light-switch-app/silabs/src/BindingHandler.cpp
arun-silabs: moved inside the actual cases

examples/light-switch-app/silabs/src/ShellCommands.cpp
arun-silabs: replaced with stol

examples/light-switch-app/silabs/include/BindingHandler.h
```diff
@@ -29,5 +29,6 @@ struct BindingCommandData
     chip::EndpointId localEndpointId = 1;
     chip::CommandId commandId;
     chip::ClusterId clusterId;
-    bool isGroup = false;
+    bool isGroup         = false;
+    uint32_t args[7];
```
mkardous-silabs, line 33: This doesn't seem like the right wat to go to pass data along. We should a Data structure which leverages unions for each command type.

examples/light-switch-app/silabs/include/LightSwitchMgr.h
```diff
@@ -38,21 +42,63 @@ class LightSwitchMgr
...
 
+        Callback mCallback = nullptr;
+        void * mContext    = nullptr;
+        bool mIsActive     = false;
+
+        osTimerId_t mHandler = nullptr;
+
+    private:
+        static void TimerCallback(void * timerCbArg);
+    };
```
mkardous-silabs, line 64: where does this come from?

examples/light-switch-app/silabs/src/BindingHandler.cpp
```diff
@@ -28,6 +28,9 @@
 
 using namespace chip;
 using namespace chip::app;
+using chip::app::Clusters::LevelControl::MoveModeEnum;
+using chip::app::Clusters::LevelControl::OptionsBitmap;
+using chip::app::Clusters::LevelControl::StepModeEnum;
```
mkardous-silabs, line 33: We shouldn't use structure / enums like this. We should stop at namespace in this case since this could cause conflicts.
```suggestion
using namespace chip::app::Clusters::LevelControl;
```

examples/light-switch-app/silabs/src/BindingHandler.cpp
```diff
@@ -85,6 +88,215 @@ void ProcessOnOffGroupBindingCommand(CommandId commandId, const EmberBindingTabl
     }
 }
 
+void ProcessLevelControlUnicastBindingCommand(BindingCommandData * data, const EmberBindingTableEntry & binding,
```
mkardous-silabs, line 91: static_cast on every argument is not the right way to do this. This argument types need to be set inside the structure as mentionned earlier.

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
```diff
@@ -26,18 +26,131 @@
...
+
+#define LONG_PRESS_TIMEOUT 3000
+#define STEP_SIZE 1
+#define TRANSITION_TIME 0
+#define OPTIONS_MASK 0
+#define OPTIONS_OVERRIDE 0
 
 LightSwitchMgr LightSwitchMgr::sSwitch;
 
+LightSwitchMgr::LightSwitchMgr() {}
```
mkardous-silabs, line 52:
```suggestion
```

examples/light-switch-app/silabs/include/LightSwitchMgr.h
```diff
@@ -38,21 +42,63 @@ class LightSwitchMgr
...
 
     void GenericSwitchOnInitialPress();
     void GenericSwitchOnShortRelease();
 
     void TriggerLightSwitchAction(LightSwitchAction action, bool isGroupCommand = false);
+    void TriggerLevelControlAction(uint8_t stepMode, bool isGroupCommand = false);
 
     static LightSwitchMgr & GetInstance() { return sSwitch; }
 
+    LightSwitchMgr();
```
mkardous-silabs, line 75: The constructor should be the first function in the section  after the structures and types.
```suggestion
    LightSwitchMgr() = default;
```

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
```diff
@@ -26,18 +26,131 @@
...
+
+AppEvent CreateNewEvent(AppEvent::AppEventTypes type)
+{
+    AppEvent aEvent;
+    aEvent.Type                      = type;
+    aEvent.Handler                   = LightSwitchMgr::GeneralEventHandler;
+    LightSwitchMgr * lightSwitch         = static_cast<LightSwitchMgr *>(&LightSwitchMgr::GetInstance());
```
mkardous-silabs, line 59: This applies to other locations as well
```suggestion
    LightSwitchMgr * lightSwitch         = &LightSwitchMgr::GetInstance();
```

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
```diff
@@ -125,6 +240,22 @@ void LightSwitchMgr::TriggerLightSwitchAction(LightSwitchAction action, bool isG
     DeviceLayer::PlatformMgr().ScheduleWork(SwitchWorkerFunction, reinterpret_cast<intptr_t>(data));
 }
 
+void LightSwitchMgr::TriggerLevelControlAction(uint8_t stepMode, bool isGroupCommand)
```
mkardous-silabs, line 243: I'm not sure i understand what we are trying to achieve - is this a long press move level ?

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
```diff
@@ -26,18 +26,131 @@
...
 using namespace chip::app;
 using namespace chip::app::Clusters;
+using namespace chip::DeviceLayer;
+using namespace chip::DeviceLayer::Silabs;
+
+#define LONG_PRESS_TIMEOUT 3000
+#define STEP_SIZE 1
+#define TRANSITION_TIME 0
+#define OPTIONS_MASK 0
+#define OPTIONS_OVERRIDE 0
```
mkardous-silabs, line 48: all these should be an in anounymous namespace as constexpr with the correct types.

mkardous-silabs (changes requested):

examples/light-switch-app/silabs/include/BindingHandler.h
arun-silabs: Added a struct for all the types

examples/light-switch-app/silabs/src/BindingHandler.cpp
arun-silabs: made this change

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
arun-silabs: removed

examples/light-switch-app/silabs/include/LightSwitchMgr.h
arun-silabs: applied the suggestion

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
arun-silabs: modified it in this file

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
arun-silabs: made these as constexpr with correct types

examples/light-switch-app/silabs/include/BindingHandler.h
```diff
@@ -19,15 +19,57 @@
...
 void BindingWorkerFunction(intptr_t context);
 
+
 struct BindingCommandData
 {
     chip::EndpointId localEndpointId = 1;
     chip::CommandId commandId;
     chip::ClusterId clusterId;
     bool isGroup = false;
+
```
mkardous-silabs, line 41: To reduce some of the duplication, you can create a Command class with the optionsMasks and optionsOverride members and have all the others structucres inherits from it.

examples/light-switch-app/silabs/include/LightSwitchMgr.h
```diff
@@ -22,7 +22,17 @@
 #include <app/util/basic-types.h>
 #include <lib/core/CHIPError.h>
 #include <lib/support/CodeUtils.h>
+#include <app/clusters/bindings/bindings.h>
+#include <platform/CHIPDeviceLayer.h>
+#include "app-common/zap-generated/ids/Clusters.h"
```
mkardous-silabs, line 27:
```suggestion
#include <app-common/zap-generated/ids/Clusters.h>
```

examples/light-switch-app/silabs/include/LightSwitchMgr.h
```diff
@@ -22,7 +22,17 @@
...
```
mkardous-silabs, line 28:
```suggestion
```

examples/light-switch-app/silabs/include/LightSwitchMgr.h
```diff
@@ -39,20 +49,69 @@ class LightSwitchMgr
         chip::EventId event;
     };
 
+    static constexpr uint8_t stepSize = 1;
+    static constexpr DataModel::Nullable<uint16_t> transitionTime = 0;
+    static constexpr chip::BitMask<OptionsBitmap> optionsMask = 0;
+    static constexpr chip::BitMask<OptionsBitmap> optionsOverride = 0;
```
mkardous-silabs, line 55: These represent a Step command, we should use the step command instead

examples/light-switch-app/silabs/src/BindingHandler.cpp
```diff
@@ -85,6 +86,263 @@ void ProcessOnOffGroupBindingCommand(CommandId commandId, const EmberBindingTabl
...
+    {
+        Clusters::LevelControl::Commands::MoveToLevel::Type moveToLevelCommand;
+        if (auto moveToLevel = std::get_if<BindingCommandData::MoveToLevel>(&data->commandData))
+        {
+            moveToLevelCommand.level           = moveToLevel->level;
+            moveToLevelCommand.transitionTime  = moveToLevel->transitionTime;
+            moveToLevelCommand.optionsMask     = moveToLevel->optionsMask;
+            moveToLevelCommand.optionsOverride = moveToLevel->optionsOverride;
+        }
+        Controller::InvokeCommandRequest(peer_device->GetExchangeManager(), peer_device->GetSecureSession().Value(), binding.remote,
```
mkardous-silabs, line 114: This should be insinde the if in case the wrong type is passed if i am not mistaken. This applies to all of the cases.

examples/light-switch-app/silabs/include/AppEvent.h
```diff
@@ -49,6 +56,10 @@ struct AppEvent
             uint8_t Action;
             int32_t Actor;
         } LightEvent;
```
mkardous-silabs, line 58: we can delete the light event for the light-switch if it is not used

examples/light-switch-app/silabs/include/LightSwitchMgr.h
```diff
@@ -39,20 +49,69 @@ class LightSwitchMgr
...
+        void * mContext    = nullptr;
+        bool mIsActive     = false;
+
+        osTimerId_t mHandler = nullptr;
+
+    private:
+        static void TimerCallback(void * timerCbArg);
+    };
+
+    LightSwitchMgr() = default;
```
mkardous-silabs, line 78: Since the LightSwitchMgr is a singleton, the constructor shouldn't be public

examples/light-switch-app/silabs/include/LightSwitchMgr.h
```diff
@@ -39,20 +49,69 @@ class LightSwitchMgr
...
+
+    static void GeneralEventHandler(AppEvent * aEvent);
+
+protected:
+
+    static void OnLongPressTimeout(Timer & timer);
+
+    Timer * mLongPressTimer = nullptr;
+    bool mDownPressed       = false;
+    bool mResetWarning      = false;
```
mkardous-silabs, line 100: This can all go in the private section since the class doesn't get inherited.

examples/light-switch-app/silabs/include/LightSwitchMgr.h
```diff
@@ -39,20 +49,69 @@ class LightSwitchMgr
...
+
 private:
     static LightSwitchMgr sSwitch;
+    void HandleLongPress();
```
mkardous-silabs, line 104: We should add comments to the fuction

examples/light-switch-app/silabs/include/LightSwitchMgr.h
```diff
@@ -39,20 +49,69 @@ class LightSwitchMgr
...
     void GenericSwitchOnShortRelease();
 
     void TriggerLightSwitchAction(LightSwitchAction action, bool isGroupCommand = false);
+    void TriggerLevelControlAction(uint8_t stepMode, bool isGroupCommand = false);
 
     static LightSwitchMgr & GetInstance() { return sSwitch; }
 
+    static void ButtonEventHandler(uint8_t button, uint8_t btnAction);
+
+    static void GeneralEventHandler(AppEvent * aEvent);
```
mkardous-silabs, line 92:
```suggestion
    static void AppEventHandler(AppEvent * aEvent);
```

examples/light-switch-app/silabs/include/LightSwitchMgr.h
```diff
@@ -39,20 +49,69 @@ class LightSwitchMgr
...
+
 private:
     static LightSwitchMgr sSwitch;
+    void HandleLongPress();
 
     static void GenericSwitchWorkerFunction(intptr_t context);
 
     chip::EndpointId mLightSwitchEndpoint   = chip::kInvalidEndpointId;
     chip::EndpointId mGenericSwitchEndpoint = chip::kInvalidEndpointId;
+    /**
```
mkardous-silabs, line 110:
```suggestion
   
     /**
```

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
```diff
@@ -26,18 +26,123 @@
...
 
 using namespace chip;
 using namespace chip::app;
 using namespace chip::app::Clusters;
+using namespace chip::DeviceLayer;
+using namespace chip::DeviceLayer::Silabs;
 
 LightSwitchMgr LightSwitchMgr::sSwitch;
 
+AppEvent CreateNewEvent(AppEvent::AppEventTypes type)
```
mkardous-silabs, line 46: this shouldn't be a global function - this needs to go in an anounymous namespace

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
```diff
@@ -26,18 +26,123 @@
...
+    aEvent.Handler                   = LightSwitchMgr::GeneralEventHandler;
+    LightSwitchMgr * lightSwitch         = &LightSwitchMgr::GetInstance();
+    aEvent.LightSwitchEvent.Context  = lightSwitch;
+    return aEvent;
+}
+
+void LightSwitchMgr::Timer::Start()
+{
+    // Starts or restarts the function timer
+    if (osTimerStart(mHandler, pdMS_TO_TICKS(100)) != osOK)
```
mkardous-silabs, line 59: This logic seems odd to force 100 as the timeout value. Can you provide mode information on the logic?

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
```diff
@@ -26,18 +26,123 @@
...
+void LightSwitchMgr::OnLongPressTimeout(LightSwitchMgr::Timer & timer)
+{
+    LightSwitchMgr * app = static_cast<LightSwitchMgr *>(timer.mContext);
+    if (app)
+    {
+        app->HandleLongPress();
+    }
+}
+
+LightSwitchMgr::Timer::Timer(uint32_t timeoutInMs, Callback callback, void * context) : mCallback(callback), mContext(context)
```
mkardous-silabs, line 103: timeoutInMs is not used in the configuration - i think the timer logic might need to be updated.

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
```diff
@@ -26,18 +26,123 @@
...
+    if (osTimerStop(mHandler) == osError)
+    {
+        SILABS_LOG("Timer stop() failed");
+        appError(CHIP_ERROR_INTERNAL);
+    }
+}
+
+void LightSwitchMgr::Timer::TimerCallback(void * timerCbArg)
+{
+    Timer * timer = static_cast<Timer *>(timerCbArg);
```
mkardous-silabs, line 139:
```suggestion
    Timer * timer = reinterpret_cast<Timer *>(timerCbArg);
```

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
mkardous-silabs: I'm not sure i agree with the idea of adding another functionnality to the button event handler...
@jmartinez-silabs do you think it is worth while to implement a "toogle" feature on button 1?

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
```diff
@@ -125,6 +232,26 @@ void LightSwitchMgr::TriggerLightSwitchAction(LightSwitchAction action, bool isG
...
+    data->commandId = LevelControl::Commands::StepWithOnOff::Id;
+    data->commandData = BindingCommandData::Step{};
+    if (auto *step = std::get_if<BindingCommandData::Step>(&data->commandData))
+    {
+        step->stepMode = static_cast<Clusters::LevelControl::StepModeEnum>(stepMode);
+        step->stepSize = LightSwitchMgr::stepSize;
+        step->transitionTime = LightSwitchMgr::transitionTime;
+        step->optionsMask = LightSwitchMgr::optionsMask;
+        step->optionsOverride = LightSwitchMgr::optionsOverride;
+    }
```
mkardous-silabs, line 250: i think this can be simplified - i'm not sure we need if check

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
```diff
@@ -159,3 +286,80 @@ void LightSwitchMgr::GenericSwitchWorkerFunction(intptr_t context)
 
     Platform::Delete(data);
 }
+
+void LightSwitchMgr::ButtonEventHandler(uint8_t button, uint8_t btnAction)
+{
+    AppEvent event = {};
+    if (btnAction == static_cast<uint8_t>(SilabsPlatform::ButtonAction::ButtonPressed))
```
mkardous-silabs, line 293:
```suggestion
    if (btnAction == to_underlying(SilabsPlatform::ButtonAction::ButtonPressed))
```

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
```diff
@@ -159,3 +286,80 @@ void LightSwitchMgr::GenericSwitchWorkerFunction(intptr_t context)
 
     Platform::Delete(data);
 }
+
+void LightSwitchMgr::ButtonEventHandler(uint8_t button, uint8_t btnAction)
```
mkardous-silabs, line 290: Can you provide information on the logic of the function?

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
```diff
@@ -125,6 +232,26 @@ void LightSwitchMgr::TriggerLightSwitchAction(LightSwitchAction action, bool isG
     DeviceLayer::PlatformMgr().ScheduleWork(SwitchWorkerFunction, reinterpret_cast<intptr_t>(data));
 }
 
+void LightSwitchMgr::TriggerLevelControlAction(uint8_t stepMode, bool isGroupCommand)
```
mkardous-silabs, line 235: The signature should be the type that is used we don't need the double conversion.

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
arun-silabs: I've modified the button logic so that the level of the bounded app can be controlled by using buttons, similar to how we do tilt and lift operations in window app on button press events

examples/light-switch-app/silabs/include/BindingHandler.h
arun-silabs: made this change

examples/light-switch-app/silabs/include/LightSwitchMgr.h
arun-silabs: Used step command

examples/light-switch-app/silabs/include/AppEvent.h
arun-silabs: deleted

examples/light-switch-app/silabs/include/LightSwitchMgr.h
arun-silabs: Moved it to private

examples/light-switch-app/silabs/include/LightSwitchMgr.h
arun-silabs: Moved to private section

examples/light-switch-app/silabs/include/LightSwitchMgr.h
arun-silabs: applied the suggestion

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
arun-silabs: Moved this to LightSwitchMgr class

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
arun-silabs: Yes, removed the if check and directly assigning the commandData

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
arun-silabs: changed the signature

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
arun-silabs: I have picked up the timer logic from window-app.

examples/light-switch-app/silabs/include/LightSwitchMgr.h
arun-silabs: Added a short description of what the function does..

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
arun-silabs: Added the brief in LightSwitchMgr.h file

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
jmartinez-silabs: @mkardous-silabs Yeah we have something similar with window-app button.
If it is documented what the buttons do.
I'm ok with implementing the long press on both buttons

examples/light-switch-app/silabs/src/ShellCommands.cpp
jmartinez-silabs: we can do it in future updates

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
```diff
@@ -159,3 +284,80 @@ void LightSwitchMgr::GenericSwitchWorkerFunction(intptr_t context)
...
+{
+    switch(aEvent->Type)
+    {
+    case AppEvent::kEventType_UpPressed:
+        LightSwitchMgr::GetInstance().TriggerLevelControlAction(LevelControl::StepModeEnum::kUp);
+        LightSwitchMgr::GetInstance().GenericSwitchOnInitialPress();
+        break;
+    case AppEvent::kEventType_UpReleased:
+        LightSwitchMgr::GetInstance().GenericSwitchOnShortRelease();
+        break;
```
jmartinez-silabs, line 356: where do you do TriggerLightSwitchAction action now? I think it is missing and if you add it here in UpPressed like to be equivalent of what we are before. You will toggle the lighting everytime you want to raise the level. 


I think the long press logic need to be used for Level Up and Down. 
Like if hold the button (after a small threshold it start moving the level (up or down) Until you release the button.
and when it is a longPress Up release you wouldn't call GenericSwitchOnShortRelease

arun-silabs: On button press, i'm using the logic of step-with-on-off. So, when the light is at min_level (OnOff value is off) and Button1 is pressed (to increase the level), OnOff value of bounded light app will also be changed. We will not need the TriggerLightSwitchAction explicitly.

Coming to the long press logic, should we configure both the actions when button-0 is long pressed - Factory-reset and level-down?

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
```diff
@@ -345,18 +345,28 @@ void LightSwitchMgr::AppEventHandler(AppEvent * aEvent)
...
+        }
         break;
     case AppEvent::kEventType_UpReleased:
         LightSwitchMgr::GetInstance().GenericSwitchOnShortRelease();
         break;
+#if 0
+    // TODO: Fix the button handling for the btn0 and btn1
     case AppEvent::kEventType_DownPressed:
-        LightSwitchMgr::GetInstance().TriggerLevelControlAction(LevelControl::StepModeEnum::kDown);
+        LightSwitchMgr::GetInstance().TriggerLightSwitchAction(action);
```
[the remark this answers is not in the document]
chirag-silabs: Done

jmartinez-silabs (approved):

examples/light-switch-app/silabs/include/LightSwitchMgr.h
mkardous-silabs: Can we use the standard way of comments?See `SwitchActionEventHandler`

examples/light-switch-app/silabs/src/LightSwitchMgr.cpp
mkardous-silabs: Can we create a follow up ticket to finish cleaning this up?

PR #149, opened 2024-11-29
[SL-UP] Add supported wifi bands attribute for lock-app.

arun-silabs:
**JIRA** : [MATTER-4302](https://jira.silabs.com/browse/MATTER-4302)

**Issue** : SupportedWifiBands attribute is not added in lock-app zap file, so we are getting UNSUPPORTED_ATTRIBUTE error while trying to read through chip-tool.

**Fix** : Added the missing attribute in zap file.


mkardous-silabs:
@arun-silabs How does the attribute get poppulated with the correct value? What does the attribute return with these changes?

PR #150, opened 2024-12-02
[SL-TEMP] Updated endpoint for the Identify cluster from endpoint 0 to endpoint 1

shgutte:
Problem: Observing unsupported cluster for IDENTIFY cluster on endpoint 1 
Solution: Updating the endpoint of the IDENTIFY cluster from 0 to 1 
Testing Done: 917 SoC


mkardous-silabs:
@shgutte Did you check if the other endpoints need to have the indentify cluster based on the specification requirements?

jmartinez-silabs:
@mkardous-silabs see https://github.com/project-chip/connectedhomeip/pull/36373 I fixed this on CSA. 
It was originally a SL-UP on release 2.4-1.4. When doing the upstream, Andrei pointed out to that the Identify cluster can't be on endpoint 0. It must be on the device-type endpoint.

Actually, sorry @shgutte , this is a SL-TEMP. It won't be carried out anywhere else.

jmartinez-silabs (approved):

mkardous-silabs (approved):

PR #154, opened 2024-12-04
[SL-UP] Adds fix for Sleep Manager on RS9116

rosahay-silabs:
- [x] Adds compilation fix for `SL_ICD_ENABLED`  and `CHIP_CONFIG_ENABLE_ICD_SERVER` as `0` (NO-JIRA)
- [x] Adds fix for Commissioning failure during Wi-Fi Network Enable (MATTER-4407)


src/platform/silabs/rs911x/BLEManagerImpl.cpp
```diff
@@ -770,7 +764,7 @@ CHIP_ERROR BLEManagerImpl::StopAdvertising(void)
     {
         // Since DriveBLEState is not called the device is still advertising
         status = rsi_ble_stop_advertising();
-        if (status != RSI_SUCCESS)
```
rosahay-silabs, line 773: Not really sure why a blanket error status is used to process the stop advertising event?
```
#define SL_STATUS_SI91X_COMMAND_DISALLOWED            (0x4E0C)
if (status == RSI_SUCCESS || status == 0x4E0C)
```

mkardous-silabs:
@rosahay-silabs i think this PR should be an sl-up and not an sl-only

src/platform/silabs/wifi/icd/WifiSleepManager.cpp
```diff
@@ -32,14 +32,14 @@ namespace {
  */
 CHIP_ERROR ConfigureLIBasedSleep()
 {
+    VerifyOrReturnError(ConfigureBroadcastFilter(true) == SL_STATUS_OK, CHIP_ERROR_INTERNAL,
+                        ChipLogError(DeviceLayer, "Failed to configure broadcasts filter."));
+
```
mkardous-silabs, line 37: Why the switch of order?

src/platform/silabs/wifi/icd/WifiSleepManager.cpp
```diff
@@ -158,9 +158,11 @@ CHIP_ERROR WifiSleepManager::VerifyAndTransitionToLowPowerMode()
     return ConfigureDTIMBasedSleep();
 
 #elif RS911X_WIFI // rs9116
+#if CHIP_CONFIG_ENABLE_ICD_SERVER
     VerifyOrReturnError(ConfigurePowerSave() == SL_STATUS_OK, CHIP_ERROR_INTERNAL);
+#endif // CHIP_CONFIG_ENABLE_ICD_SERVER
```
mkardous-silabs, line 163: the #if isn't necessary - WifiSleepManager is only included for icd builds.

src/platform/silabs/wifi/rs911x/WifiInterface.cpp
```diff
@@ -776,6 +769,22 @@ void ProcessEvent(WifiEvent event)
...
+ * @return sl_status_t Returns underlying Wi-Fi initialization error
+ */
+sl_status_t sl_matter_wifi_platform_init(void)
+{
+    uint32_t rsi_status = sl_matter_wifi_init();
+    if (rsi_status != RSI_SUCCESS)
+    {
+        ChipLogError(DeviceLayer, "sl_matter_wifi_platform_init: sl_matter_wifi_init failed: %ld", rsi_status);
+        return SL_STATUS_FAIL;
+    }
```
mkardous-silabs, line 784: This can be simplified with a verify or return statement

src/platform/silabs/wifi/icd/WifiSleepManager.cpp
rosahay-silabs: It is because the API will fail if invoked after going to sleep and it is aligned with what is provided in the sample application

src/platform/silabs/wifi/icd/WifiSleepManager.cpp
rosahay-silabs: If we set `CHIP_CONFIIG_ENABLE_ICD_SERVER` to 0 or undefine it, the sleep manager still gets included right? I know the component if not used will not include this.

rosahay-silabs:
> @rosahay-silabs i think this PR should be an sl-up and not an sl-only

Yes, initially I thought it would a temporary workaround as was done on 917 NCP. changing it to SL-UP

src/platform/silabs/rs911x/BLEManagerImpl.cpp
```diff
@@ -731,7 +721,10 @@ CHIP_ERROR BLEManagerImpl::StartAdvertising(void)
 
 exit:
     // TODO: Add MapBLEError to return the correct error code
```
mkardous-silabs, line 723: Is this TODO still necessary? If so, can we apply the change?

src/platform/silabs/wifi/icd/WifiSleepManager.cpp
mkardous-silabs: I don't have an issue with switching them but Selective Listening works with the current implementation so in seems to work in practice; at least for the 917 SoC. 

Is the change specific due to limitations of the 9116?

src/platform/silabs/wifi/icd/WifiSleepManager.cpp
mkardous-silabs: No, the sleep manager is only included for now on ICD builds is gatted by the build flag or the component. This isn't necessary.

src/platform/silabs/wifi/rs911x/WifiInterface.cpp
```diff
@@ -776,6 +769,18 @@ void ProcessEvent(WifiEvent event)
     }
 }
 
+/**
+ * @brief Wifi initialization called from app main
+ *
+ * @return sl_status_t Returns underlying Wi-Fi initialization error
+ */
```
mkardous-silabs, line 776: Since there can be multiple implementations of the same function for the different paltforms, the comment should go in the header for this function.

src/platform/silabs/wifi/rs911x/WifiInterface.cpp
```diff
@@ -776,6 +769,18 @@ void ProcessEvent(WifiEvent event)
...
+sl_status_t sl_matter_wifi_platform_init(void)
+{
+    uint32_t rsi_status = sl_matter_wifi_init();
+    VerifyOrReturnError(rsi_status == RSI_SUCCESS, SL_STATUS_FAIL);
```
mkardous-silabs, line 780: This can be combined in a single statement.

src/platform/silabs/wifi/icd/WifiSleepManager.cpp
rosahay-silabs: I would think it is better and cleaner to invoke all the operations before going to sleep.

I guess 917 SoC does not show this problem, technically the API should be configurable even after sleep has been invoked but in practice NCP combinations are having the issue. This included 917 NCP.

src/platform/silabs/wifi/icd/WifiSleepManager.cpp
rosahay-silabs: I am not sure what I am doing wrong, but can you check by setting the following options in `matter_icd_core.slcc`
```yaml
- name: CHIP_CONFIG_ENABLE_ICD_SERVER
  value: 0
- name: SL_ICD_ENABLED
  value: 0
- name: SL_SI91X_NPSS_GPIO_BTN_HANDLER
  value: 0
```

src/platform/silabs/rs911x/BLEManagerImpl.cpp
rosahay-silabs: this TODO might still be necessary, having said that let's put it across in another PR since this PR is focused on the sleep issue.

src/platform/silabs/wifi/icd/WifiSleepManager.cpp
mkardous-silabs: Can we open follow up issues with the wiseconnect team to review this? The documentation doesn't specify and there should be no API difference between SoC and NCP.

Can you add a comment in the comment to specify that the sleep command needs to be last thing called?

src/platform/silabs/wifi/icd/WifiSleepManager.cpp
mkardous-silabs: Why would we disable the ICD features in the icd_core component? I'm not sure i understand what we are trying to test or achieve here...

src/platform/silabs/rs911x/BLEManagerImpl.cpp
mkardous-silabs: Can you open an task under this story that we need to fix this?
https://jira.silabs.com/browse/MATTER-4426

rosahay-silabs: https://jira.silabs.com/browse/MATTER-4426

src/platform/silabs/wifi/icd/WifiSleepManager.cpp
```diff
@@ -32,14 +32,14 @@ namespace {
  */
 CHIP_ERROR ConfigureLIBasedSleep()
 {
+    VerifyOrReturnError(ConfigureBroadcastFilter(true) == SL_STATUS_OK, CHIP_ERROR_INTERNAL,
+                        ChipLogError(DeviceLayer, "Failed to configure broadcasts filter."));
+    // powersave has to be invoked at the last else the configure boardcast filter fails
```
mkardous-silabs, line 37:
```suggestion
    
    // Allowing the device to go to sleep must be the last actions to avoid configuration failures.
```

src/platform/silabs/wifi/icd/WifiSleepManager.cpp
```diff
@@ -51,12 +51,12 @@ CHIP_ERROR ConfigureLIBasedSleep()
  */
 CHIP_ERROR ConfigureDTIMBasedSleep()
 {
-    VerifyOrReturnError(ConfigurePowerSave(RSI_SLEEP_MODE_2, ASSOCIATED_POWER_SAVE, 0) == SL_STATUS_OK, CHIP_ERROR_INTERNAL,
-                        ChipLogError(DeviceLayer, "Failed to enable to enable DTIM basedsleep."));
-
     VerifyOrReturnError(ConfigureBroadcastFilter(false) == SL_STATUS_OK, CHIP_ERROR_INTERNAL,
                         ChipLogError(DeviceLayer, "Failed to configure broadcast filter."));
 
+    VerifyOrReturnError(ConfigurePowerSave(RSI_SLEEP_MODE_2, ASSOCIATED_POWER_SAVE, 0) == SL_STATUS_OK, CHIP_ERROR_INTERNAL,
```
mkardous-silabs, line 57:
```suggestion
    // Allowing the device to go to sleep must be the last actions to avoid configuration failures.
    VerifyOrReturnError(ConfigurePowerSave(RSI_SLEEP_MODE_2, ASSOCIATED_POWER_SAVE, 0) == SL_STATUS_OK, CHIP_ERROR_INTERNAL,
```

mkardous-silabs (approved):

jmartinez-silabs (approved):

PR #158, opened 2024-12-05
[SL-UP] Provision: Dynamic buffer allocation

rcasallas-silabs:
Now the provision.py script allows to set the size of the data buffer with the --buffer_size argument. This requires changes on the provision library.

Tested on BRD4187C.

### DIFF
--- a/examples/platform/silabs/provision/ProvisionStorageDefault.cpp
+++ b/examples/platform/silabs/provision/ProvisionStorageDefault.cpp
@@ -70,15 +70,25 @@ CHIP_ERROR ErasePage(uint32_t addr)
     return chip::DeviceLayer::Silabs::GetPlatform().FlashErasePage(addr);
 }
 
-CHIP_ERROR WritePage(uint32_t addr, const uint8_t * data, size_t size)
-{
-    return chip::DeviceLayer::Silabs::GetPlatform().FlashWritePage(addr, data, size);
-}
-
 size_t RoundNearest(size_t n, size_t multiple)
 {
     return (n % multiple) > 0 ? n + (multiple - n % multiple) : n;
 }
+
+CHIP_ERROR WritePage(uint32_t addr, const uint8_t * data, size_t size, size_t max)
+{
+    size_t size_32 = RoundNearest(size, 4);
+    if(size_32 != size)
+    {
+        uint8_t * p = (uint8_t *) data;
+        VerifyOrReturnError(size_32 <= max, CHIP_ERROR_BUFFER_TOO_SMALL);
+        memset(p + size, 0xff, size_32 - size);
+        size = size_32;
+    }
+    return chip::DeviceLayer::Silabs::GetPlatform().FlashWritePage(addr, data, size);
+}
+
+
 
 CHIP_ERROR WriteFile(Storage & store, SilabsConfig::Key offset_key, SilabsConfig::Key size_key, const ByteSpan & value)
 {
@@ -89,17 +99,7 @@ CHIP_ERROR WriteFile(Storage & store, SilabsConfig::Key offset_key, SilabsConfig
         ReturnErrorOnFailure(ErasePage(base_addr));
     }
 
-    memcpy(Storage::aux_buffer, value.data(), value.size());
-    if (value.size() < Storage::kArgumentSizeMax)
-    {
-        memset(Storage::aux_buffer + value.size(), 0xff, Storage::kArgumentSizeMax - value.size());
-    }
-
-    ChipLogProgress(DeviceLayer, "WriteFile, addr:0x%06x+%03u, size:%u", (unsigned) base_addr, (unsigned) sCredentialsOffset,
-                    (unsigned) value.size());
-    // ChipLogByteSpan(DeviceLayer, ByteSpan(value.data(), value.size() < kDebugLength ? value.size() : kDebugLength));
-
-    ReturnErrorOnFailure(WritePage(base_addr + sCredentialsOffset, Storage::aux_buffer, Storage::kArgumentSizeMax));
+    ReturnErrorOnFailure(WritePage(base_addr + sCredentialsOffset, value.data(), value.size(), store.GetBufferSize()));
 
     // Store file offset
     ReturnErrorOnFailure(SilabsConfig::WriteConfigValue(offset_key, (uint32_t) sCredentialsOffset));
@@ -119,7 +119,6 @@ CHIP_ERROR ReadFileByOffset(Storage & store, const char * description, uint32_t 
     ByteSpan span(address, size);
     ChipLogProgress(DeviceLayer, "%s, addr:0x%06x+%03u, size:%u", description, (unsigned) base_addr, (unsigned) offset,
                     (unsigned) size);
-    // ChipLogByteSpan(DeviceLayer, ByteSpan(span.data(), span.size() < kDebugLength ? span.size() : kDebugLength));
     return CopySpanToMutableSpan(span, value);
 }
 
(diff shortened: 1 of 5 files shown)


src/platform/silabs/provision/ProvisionEncoder.h
```diff
@@ -20,24 +20,47 @@
...
         this->in    = at_end ? end : begin;
         this->out   = ptr;
     }
+    void Finish()
+    {
+        if(this->begin && allocated)
+        {
+            free(this->begin);
+        }
+        this->begin = this->end = this->in = this->out = nullptr;
```
mkardous-silabs, line 62: I'm not sure i follow the changes here - We only seem to free in the first block of the chain. Is there a risk of a memory leak here?

src/platform/silabs/provision/ProvisionStorage.h
```diff
@@ -60,7 +60,8 @@ enum ID : uint16_t
     kGeneratorFW   = 0x0135,
     kProductionFW  = 0x0136,
     kCertToolPath  = 0x0137,
-    kPylinkLib     = 0x0138,
+    kPylinkLib     = 0x013a,
```
mkardous-silabs, line 63: Is it normal that the pylink lib value changes?

src/platform/silabs/provision/ProvisionStorage.h
```diff
@@ -160,9 +160,7 @@ struct Storage : public GenericStorage,
     static constexpr size_t kTotalPayloadDataSize = kTotalPayloadDataSizeInBits / 8;
 
 public:
-    static uint8_t aux_buffer[Storage::kArgumentSizeMax];
-
-    friend class Manager;
+   friend class Manager;
```
mkardous-silabs, line 163: I'm not sure i understand the necessity of this change.

src/platform/silabs/provision/ProvisionEncoder.h
```diff
@@ -20,24 +20,47 @@
...
 #include <stdint.h>
+#include <cstdlib>
 
 namespace chip {
 namespace DeviceLayer {
 namespace Silabs {
 namespace Provision {
 namespace Encoding {
 
+/*
```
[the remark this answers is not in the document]
rcasallas-silabs: I have added some context.

jmartinez-silabs (approved):

examples/platform/silabs/provision/ProvisionStorageDefault.cpp
```diff
@@ -70,16 +70,26 @@ CHIP_ERROR ErasePage(uint32_t addr)
...
-    return chip::DeviceLayer::Silabs::GetPlatform().FlashWritePage(addr, data, size);
+    return (n % multiple) > 0 ? n + (multiple - n % multiple) : n;
 }
 
-size_t RoundNearest(size_t n, size_t multiple)
+CHIP_ERROR WritePage(uint32_t addr, const uint8_t * data, size_t size, size_t max)
 {
-    return (n % multiple) > 0 ? n + (multiple - n % multiple) : n;
+    size_t size_32 = RoundNearest(size, 4);
+    if(size_32 != size)
```
lpbeliveau-silabs, line 81:
```suggestion
    // If the input data is smaller than a flash page, pad the input buffer with "0xff"
    if(size_32 != size)
```

examples/platform/silabs/provision/ProvisionStorageDefault.cpp
```diff
@@ -70,16 +70,26 @@ CHIP_ERROR ErasePage(uint32_t addr)
...
+    {
+        uint8_t * p = (uint8_t *) data;
+        VerifyOrReturnError(size_32 <= max, CHIP_ERROR_BUFFER_TOO_SMALL);
```
lpbeliveau-silabs, line 84: So essentially this error means that the buffer is smaller than a page. Seems like it would be more explicit to check that from the start, since this method is Called _WritePage_

examples/platform/silabs/provision/ProvisionStorageDefault.cpp
```diff
@@ -70,16 +70,26 @@ CHIP_ERROR ErasePage(uint32_t addr)
...
 {
-    return chip::DeviceLayer::Silabs::GetPlatform().FlashWritePage(addr, data, size);
+    return (n % multiple) > 0 ? n + (multiple - n % multiple) : n;
 }
 
-size_t RoundNearest(size_t n, size_t multiple)
+CHIP_ERROR WritePage(uint32_t addr, const uint8_t * data, size_t size, size_t max)
 {
-    return (n % multiple) > 0 ? n + (multiple - n % multiple) : n;
+    size_t size_32 = RoundNearest(size, 4);
```
lpbeliveau-silabs, line 80: This implies that this is the size rounded in _uint32_, while the name suggest that we are writing a whole page.

My comment below might be wrong because I see that the logic doesn't seem to indicate that `multiple` matches the page size... which is why this is in dire need of commenting.

src/platform/silabs/efr32/CHIPCryptoPALPsaEfr32.cpp
```diff
@@ -43,6 +43,19 @@
 #include <string.h>
 #include <type_traits>
 
+// Includes needed for certificate parsing
```
lpbeliveau-silabs, line 46: I am assuming this is mbedtls code since I have seen some of this elsewhere in the repo, so I am skipping it. Am I right?

If so, since this lives in our files, maybe it would be worthy to identify the parts that are EFR32 specific? To shorten the review time and to allow for someone else to pickup this file in the future if need be?

examples/platform/silabs/provision/ProvisionStorageDefault.cpp
rcasallas-silabs: Suggestion applied.

src/platform/silabs/efr32/CHIPCryptoPALPsaEfr32.cpp
rcasallas-silabs: These functions are actually part of the CHIPCryptoPALPSA.h interface. They were implemented in the old CHIPCryptoPALPsaEfr32.cpp, but are missing in CHIPCryptoPALPSA.cpp. As far I know, devices don't use them, but the test driver fails to compile without them. I suspect there is an error in the GN dependencies, but fixing them would expand the scope of this PR.

examples/platform/silabs/provision/ProvisionStorageDefault.cpp
rcasallas-silabs: You were not far-off. The size is not rounded to the page size, but the nearest 32-bit boundary. Otherwise the flash driver fails to write. I have updated the comment accodingly.

examples/platform/silabs/provision/ProvisionStorageDefault.cpp
```diff
@@ -70,16 +70,26 @@ CHIP_ERROR ErasePage(uint32_t addr)
...
 
-CHIP_ERROR WritePage(uint32_t addr, const uint8_t * data, size_t size)
+size_t RoundNearest(size_t n, size_t multiple)
 {
-    return chip::DeviceLayer::Silabs::GetPlatform().FlashWritePage(addr, data, size);
+    return (n % multiple) > 0 ? n + (multiple - n % multiple) : n;
 }
 
-size_t RoundNearest(size_t n, size_t multiple)
+CHIP_ERROR WritePage(uint32_t addr, const uint8_t * data, size_t size, size_t max)
```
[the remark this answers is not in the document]
rcasallas-silabs: Commented.

PR #174, opened 2024-12-09
[SL-ONLY] Add custom rules to selective listening

mkardous-silabs:
##### Description
PR fixes Selective Listening to work with ecosystems with uniques behavior; in this case, the homdepod mini.
The homepod has two particular devices:

- It commissionnes a second fabric without oppening a commissioning window - only with the Fail Safe Timer feature
- It commissionnes two fabrics with a single subscriptions

To fix the first limitation, we register the ApplicationSleepManager with the ICDManager to get notified of transition from ActiveMode to IdleMode. There is no other mechanism at the moment to know when the fail safe timer gets armed or disarmed. In a follow up, the Application SleepManager can be simplified since it covers the commissioning as well.

To fix the second issue, the PR adds a custom rule set to to cover special use cases that would prevent the device from going to LI based sleep. In the case of apple, we do extra checks on the fabrics to validate that we can go to LI based sleep even if we only have a single subscription for two fabrics.
In a follow up, we should introduce a rule template that can be implemented by different "rules engine" to simplfy the maintenance and adding new checks. The way is done right now has a lot of overhead.

#### Tests
Manual tests with apple


examples/platform/silabs/wifi/icd/ApplicationSleepManager.cpp
```diff
@@ -92,24 +103,97 @@ bool ApplicationSleepManager::CanGoToLIBasedSleep()
...
+                }
+                else
+                {
+                    // Fabric doesn't meet the requirements to allow us to go to LI based sleep
+                    break;
+                }
+            }
+            else
+            {
+                ChipLogProgress(AppServer, "Fabric index %u has an active subscriptions!", it->GetFabricIndex());
```
jmartinez-silabs, line 134: The logs in this function seem too verbose to me.

examples/platform/silabs/wifi/icd/ApplicationSleepManager.cpp
```diff
@@ -92,24 +103,97 @@ bool ApplicationSleepManager::CanGoToLIBasedSleep()
...
+{
+    bool canGoToLIBasedSleep = false;
+    switch (to_underlying(vendorId))
+    {
+    case to_underlying(SpecialCaseVendorID::kAppleKeychain):
+        canGoToLIBasedSleep = ProcessAppleKeychainEdgeCase();
+        break;
+
+    default:
+        ChipLogProgress(AppServer, "Vendor ID doesn't have an edge case to process.");
```
jmartinez-silabs, line 152: I would remove this too In most case this will get printed

examples/platform/silabs/wifi/icd/ApplicationSleepManager.cpp
```diff
@@ -92,24 +103,97 @@ bool ApplicationSleepManager::CanGoToLIBasedSleep()
...
-            ChipLogProgress(AppServer, "Fabric index %u has an active subscriptions!", it->GetFabricIndex());
         }
     }
 
     return canGoToLIBasedSleep;
 }
 
+bool ApplicationSleepManager::ProcessSpecialVendorIDCase(chip::VendorId vendorId)
+{
+    bool canGoToLIBasedSleep = false;
```
jmartinez-silabs, line 144: This boolean name only makes sense in the context of how you use this function but not within the function itself. In case this function has other use later on, I would rename this

examples/platform/silabs/wifi/icd/ApplicationSleepManager.cpp
```diff
@@ -92,24 +103,97 @@ bool ApplicationSleepManager::CanGoToLIBasedSleep()
...
+
+    default:
+        ChipLogProgress(AppServer, "Vendor ID doesn't have an edge case to process.");
+        break;
+    }
+
+    return canGoToLIBasedSleep;
+}
+
+bool ApplicationSleepManager::ProcessAppleKeychainEdgeCase()
```
jmartinez-silabs, line 159: i'd recommend removing the "Apple" from the func name

examples/platform/silabs/wifi/icd/ApplicationSleepManager.cpp
```diff
@@ -92,24 +103,97 @@ bool ApplicationSleepManager::CanGoToLIBasedSleep()
...
+{
+    bool canGoToLIBasedSleep = false;
```
jmartinez-silabs, line 161: same comment about the var name

src/platform/silabs/CHIPPlatformConfig.h
```diff
@@ -118,6 +118,8 @@
 #define CHIP_CONFIG_ICD_CLIENTS_SUPPORTED_PER_FABRIC SL_ICD_SUPPORTED_CLIENTS_PER_FABRIC
 #endif // CHIP_CONFIG_ICD_CLIENTS_SUPPORTED_PER_FABRIC
 
+#define CHIP_CONFIG_ICD_OBSERVERS_POOL_SIZE 3
```
jmartinez-silabs, line 121: Suggestion so the Userconfig can still override this to increase it more
```suggestion
#ifndef CHIP_CONFIG_ICD_OBSERVERS_POOL_SIZE
#define CHIP_CONFIG_ICD_OBSERVERS_POOL_SIZE 3
#endif
```

jmartinez-silabs (approved):

src/platform/silabs/CHIPPlatformConfig.h
```diff
@@ -118,6 +118,11 @@
 #define CHIP_CONFIG_ICD_CLIENTS_SUPPORTED_PER_FABRIC SL_ICD_SUPPORTED_CLIENTS_PER_FABRIC
 #endif // CHIP_CONFIG_ICD_CLIENTS_SUPPORTED_PER_FABRIC
 
+#ifndef CHIP_CONFIG_ICD_OBSERVERS_POOL_SIZE
+#define CHIP_CONFIG_ICD_OBSERVERS_POOL_SIZE 3
+#endif // CHIP_CONFIG_ICD_OBSERVERS_POOL_SIZE
+static_assert(CHIP_CONFIG_ICD_OBSERVERS_POOL_SIZE >= 3, "ICD Observers pool size must be at least 3");
```
jmartinez-silabs, line 124: Only in the wifi use case no? else it is 2. Probably need a added #ifdef

mkardous-silabs: good catch! updated

examples/platform/silabs/wifi/icd/ApplicationSleepManager.cpp
```diff
@@ -92,24 +103,92 @@ bool ApplicationSleepManager::CanGoToLIBasedSleep()
...
+bool ApplicationSleepManager::ProcessKeychainEdgeCase()
+{
+    bool hasValidException = false;
+
+    for (auto it = mFabricTable->begin(); it != mFabricTable->end(); ++it)
+    {
+        if (to_underlying(it->GetVendorId()) == to_underlying(SpecialCaseVendorID::kAppleHome) &&
+            mSubscriptionsInfoProvider->FabricHasAtLeastOneActiveSubscription(it->GetFabricIndex()))
+        {
+            hasValidException = true;
```
lpbeliveau-silabs, line 163:
```suggestion
            hasValidException = true;
            break;
```

examples/platform/silabs/wifi/icd/ApplicationSleepManager.h
```diff
@@ -131,20 +133,54 @@ class ApplicationSleepManager : public chip::app::ReadHandler::ApplicationCallba
...
+     * subcription. Applye Keychain fabric acts as a safety and doesn't have an active fabric with the device. As such, we need an
+     * alternate method to check if the device can go to LI based sleep.
+     *
+     * This method checks if there is any fabric with the Apple Home vendor ID that
+     * has at least one active subscription. If such a fabric is found, it allows
+     * the device to go to LI based sleep.
+     *
+     * @return true if the Apple Keychain edge case allows low-power mode, false otherwise.
+     */
+    bool ProcessKeychainEdgeCase();
```
lpbeliveau-silabs, line 174: I know this is very explicit by the comment above, but how about putting _Apple_ in the method name so when we se it called in the code we don't have to come back here to know what is going on?

(Or we could have a functioning intellisense that shows the comment when we hover the function but who has this kind of technology...)

lpbeliveau-silabs (approved):

examples/platform/silabs/wifi/icd/ApplicationSleepManager.cpp
```diff
@@ -16,12 +16,23 @@
...
 
 namespace chip {
 namespace app {
 namespace Silabs {
 
+namespace {
+
+enum class SpecialCaseVendorID : uint16_t
+{
+    kAppleHome     = 4937,
```
rcasallas-silabs, line 30: This is equal to VendorId::Apple. This enumeration only adds kAppleKeychain. Could that be part of VendorId too? Sounds unlikely, but that would avoid the ugly to_underlying cast.

examples/platform/silabs/wifi/icd/ApplicationSleepManager.h
jmartinez-silabs: I made the inverse comment before (he had apple in) IMO we should not publically mention Apple in our function names

examples/platform/silabs/wifi/icd/ApplicationSleepManager.cpp
```diff
@@ -92,24 +103,92 @@ bool ApplicationSleepManager::CanGoToLIBasedSleep()
...
+    bool hasValidException = false;
+    switch (to_underlying(vendorId))
+    {
+    case to_underlying(SpecialCaseVendorID::kAppleKeychain):
+        hasValidException = ProcessKeychainEdgeCase();
+        break;
+
+    default:
+        break;
+    }
```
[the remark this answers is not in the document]
mkardous-silabs: Won't do this one for now since i'm going to refactor this at some point

examples/platform/silabs/wifi/icd/ApplicationSleepManager.cpp
```diff
@@ -92,24 +102,93 @@ bool ApplicationSleepManager::CanGoToLIBasedSleep()
...
+    return hasValidException;
+}
+
+bool ApplicationSleepManager::ProcessKeychainEdgeCase()
+{
+    bool hasValidException = false;
+
+    for (auto it = mFabricTable->begin(); it != mFabricTable->end(); ++it)
+    {
+        if (to_underlying(it->GetVendorId()) == chip::VendorId::Apple &&
```
rcasallas-silabs, line 159: Are you sure you need to_underlying? I think GetVendorId() returns chip::VendorId.
```suggestion
        if (it->GetVendorId() == chip::VendorId::Apple &&
```

PR #176, opened 2024-12-10
[SL-UP] Fix test event trigger enable validation

mkardous-silabs:
#### Description
When the SDK check if the test event trigger is enabled, it checks if the test event trigger enableKey is equal to a zero Bytespan. If the enableKey matches, returns false that the feature is not enabled.

In the current implementation, if we fail to read the enable key, we immediatly return that the key does not match. If the key we are trying to match was the zero bytespan, this triggers the stack to think the feature is enabled since the platform enable key does not match the zero bytepspan which is incorrect.

The other issue was that we return that the enableKeys don't match if `!enableKeySpan.empty() `. When checking with the zero bytespan, this returns false immediatly without checking the platform enable key,

The PR changes the logic to always check if the data matches or not which, trough the bytespan function, covers all the edge cases.

PR also does a small cleanup of the provisioning test event trigger related code.

#### Tests
Manual tests all the scenarios - we really need a way to unit test stuff in the examples directory.

### DIFF
--- a/examples/platform/silabs/SilabsTestEventTriggerDelegate.cpp
+++ b/examples/platform/silabs/SilabsTestEventTriggerDelegate.cpp
@@ -18,6 +18,7 @@
 
 #include "SilabsTestEventTriggerDelegate.h"
 #include <ProvisionManager.h>
+#include <lib/core/ErrorStr.h>
 
 using namespace ::chip::DeviceLayer;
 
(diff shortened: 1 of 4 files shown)


examples/platform/silabs/SilabsTestEventTriggerDelegate.cpp
```diff
@@ -18,21 +18,27 @@
...
+    uint8_t storedEnableKey[TestEventTriggerDelegate::kEnableKeyLength] = { 0 };
+    MutableByteSpan storedEnableKeySpan(storedEnableKey);
 
-    // Return false if we were not able to get the enableKey
-    VerifyOrReturnValue(
-        Silabs::Provision::Manager::GetInstance().GetStorage().GetTestEventTriggerKey(enableKeySpan) == CHIP_NO_ERROR, false);
+    CHIP_ERROR error = Silabs::Provision::Manager::GetInstance().GetStorage().GetTestEventTriggerKey(storedEnableKeySpan);
+    if (error != CHIP_NO_ERROR)
+    {
+        // If we fail to the read the enableKey from storage, the MutableByteSpan will be empty to a zero bytespan.
```
jmartinez-silabs, line 35:
```suggestion
        // If we fail to read the enableKey from storage, the MutableByteSpan will be empty  and not equal to a zero bytespan.
```
I'm actually not sure if the last part of the edit is true but I can't quite understand what you meant if that isn't it

mkardous-silabs: The phrase should be :

If we fail to read the enableKey from storage, the MutableByteSpan is not modified by the getter which leaves the span equal to a zero bytepsan (size = 0).

examples/platform/silabs/SilabsTestEventTriggerDelegate.cpp
```diff
@@ -18,21 +18,27 @@
...
+    MutableByteSpan storedEnableKeySpan(storedEnableKey);
 
-    // Return false if we were not able to get the enableKey
-    VerifyOrReturnValue(
-        Silabs::Provision::Manager::GetInstance().GetStorage().GetTestEventTriggerKey(enableKeySpan) == CHIP_NO_ERROR, false);
+    CHIP_ERROR error = Silabs::Provision::Manager::GetInstance().GetStorage().GetTestEventTriggerKey(storedEnableKeySpan);
+    if (error != CHIP_NO_ERROR)
+    {
+        // If we fail to read the enableKey from storage, the MutableByteSpan is not modified by the getter which leaves the span equal to a zero bytepsan (size = 0).
+        // This garantees that we will be able to inform the stack that the test event trigger is not enabled when the stack tries
```
mykrupp, line 36:
```suggestion
        // This guarantees that we will be able to inform the stack that the test event trigger is not enabled when the stack tries
```

mykrupp (approved):

PR #179, opened 2024-12-13
[SL-TEMP] Temp-fix for delay 917 NCP sleep till wifi-connection

bhmanda-silabs:
With 917 NCP sleepy apps, if device is allowed to sleep before wifi connection ,it is failing to wakeup in si91x_req_wakeup().So delayed sleep tilll wifi connection. 
Added SLI_SI91X_MCU_INTERFACE macro before verifying the device can sleep, so that NCP will not enter power save mode till wifi-connection.


mkardous-silabs:
Is there a jira ticket  on the matter side and the wifi sdk side to properly support sleep for the NCP?

mkardous-silabs (approved):

rcasallas-silabs (approved):

PR #180, opened 2024-12-13
[SL-ONLY] Bugfix/tracing log format

lpbeliveau-silabs:
Updated the formatting of the timestamps logs and added a check for empty buffer in tracing shell commands

### DIFF
--- a/src/platform/silabs/tracing/SilabsTracing.cpp
+++ b/src/platform/silabs/tracing/SilabsTracing.cpp
@@ -15,7 +15,7 @@
  *
  ******************************************************************************/
 #include "SilabsTracing.h"
-#include <cstdio> // Include the necessary header for snprintf
+#include <cstdio> // for snprintf
 #include <cstring>
 #include <lib/support/CodeUtils.h>
 #include <lib/support/PersistentData.h>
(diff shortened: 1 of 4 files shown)


src/platform/silabs/tracing/SilabsTracing.cpp
```diff
@@ -15,9 +15,11 @@
  *
  ******************************************************************************/
 #include "SilabsTracing.h"
+#include <cstdio> // Include the necessary header for snprintf
 #include <cstring>
 #include <lib/support/CodeUtils.h>
 #include <lib/support/PersistentData.h>
+#include <string> // Include the necessary header for std::string
```
jmartinez-silabs, line 22: you have `cstring`, `string` and `cstdio`

I would have expected either `stdio.h. with string.h` or `cstdio with cstring` but not three of them no?

jmartinez-silabs (approved):

src/platform/silabs/tracing/SilabsTracing.cpp
```diff
@@ -122,6 +124,22 @@ const char * OperationTypeToString(OperationType type)
...
 
+std::string FormatTimeStamp(std::chrono::milliseconds time)
+{
+    auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(time).count();
+    auto s  = ms / 1000;
+    ms %= 1000;
+    auto m = s / 60;
+    s %= 60;
+    auto h = m / 60;
+    m %= 60;
```
mkardous-silabs, line 135: le header time a des constexpr pour faciler les conversions. Ca vaudrait le coup d'au moins utiliser les constantes.

src/platform/silabs/tracing/SilabsTracing.cpp
```diff
@@ -122,6 +124,22 @@ const char * OperationTypeToString(OperationType type)
     }
 }
 
+std::string FormatTimeStamp(std::chrono::milliseconds time)
+{
+    auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(time).count();
```
mkardous-silabs, line 129: pas sur de comprendre l'utilite du duraction cast. time est deja un `std::chrono::milliseconds`.

mkardous-silabs (approved):

src/platform/silabs/tracing/SilabsTracing.cpp
lpbeliveau-silabs: cstring and cstdio are compatibility includes for C standard library to be included in C++, <string.h> is for C++ string. Let me check if there would be a clean way to ditch C++ string type here...

PR #186, opened 2024-12-16
[SL-UP] Adds changes for generator firmware

rosahay-silabs:
- Adds changes to Provision Storage flash to save private key into storage
- Adds changes to use pointer to execute Provision APIs

### DIFF
--- a/src/platform/silabs/platformAbstraction/WiseMcuSpam.cpp
+++ b/src/platform/silabs/platformAbstraction/WiseMcuSpam.cpp
@@ -39,7 +39,7 @@ extern "C" {
 #ifdef SL_CATALOG_SIMPLE_BUTTON_PRESENT
 #include "sl_si91x_button.h"
 #include "sl_si91x_button_pin_config.h"
-#endif  //SL_CATALOG_SIMPLE_BUTTON_PRESENT
+#endif // SL_CATALOG_SIMPLE_BUTTON_PRESENT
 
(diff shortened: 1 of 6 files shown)


src/platform/silabs/platformAbstraction/WiseMcuSpam.cpp
```diff
@@ -96,10 +95,10 @@ CHIP_ERROR SilabsPlatform::Init(void)
     // TODO: Setting the highest priority for SVCall_IRQn to avoid the HardFault issue
     NVIC_SetPriority(SVCall_IRQn, CORE_INTERRUPT_HIGHEST_PRIORITY);
 
-#if CHIP_CONFIG_ENABLE_ICD_SERVER == 0
+#if (CHIP_CONFIG_ENABLE_ICD_SERVER == 0) && (SL_PROVISION_GENERATOR == 0)
```
jmartinez-silabs, line 98: Correct me if I am wrong but SL_PROVISION_GENERATOR is never defined when it isn't the GFW. This could lead to an undefined value being used.

maybe 
`#if defined(CHIP_CONFIG_ENABLE_ICD_SERVER) && (CHIP_CONFIG_ENABLE_ICD_SERVER == 0) `
would do the trick. I don't think CHIP_CONFIG_ENABLE_ICD_SERVER us defined in gfw

jmartinez-silabs (approved):

chirag-silabs (approved):

PR #187, opened 2024-12-16
[SL-ONLY] re-enable channel steering for series2. update the ifdef condition

jmartinez-silabs:
re-enable channel steering for series 2. Update the ifdef condition to use `SL_CATALOG_RAIL_UTIL_IEEE802154_FAST_CHANNEL_SWITCHING_PRESENT` so it works independently with S2 or S3 and can be enabled/disable at will


selissia (approved):
Tested with concurrent CMP app without the bootloader DMA fix.

mkardous-silabs (approved):

selissia:
Sequential works too

PR #188, opened 2024-12-17
[SL-UP] Bump matter support

jmartinez-silabs:
Bump Matter_support to get the new provision libs. 
This also brings a few teaks from 917 crypto


jmartinez-silabs:
#212 deals with this already and more

PR #214, opened 2025-01-17
[SL-UP] Allowing DUT to sleep after commissioning

chirag-silabs:
**Description**
The 917 SoC DUT was going in the DTIM sleep during the commissioning process due to which CASE session error was seen during commissioning.

**Fix**
Setting the Device to High power once the commission session starts and sleepy once the commission process is done to avoid issues during the commissioning process

So there are a total of 4 state, and modifications are made as discussed:

**HIGH** -> Requesting the high performance during commissioning and during retry
**DTIM/Connected** -> Init, After commissioning is done (removing the high performance) and power cycle join
**Deepsleep** -> On Commissioning window closed if DUT isn't commissioned, disconnect
**LI** -> Subscriptions

#### Testing
Tested locally with 917 SoC, commissioning, power cycle and subscription.


jmartinez-silabs:
@chirag-silabs Fundamentally, What is causing the issue? DTIM sleep should be available during commissioning. We would like to get an explanation on why it is not possible with our wifi platform.

chirag-silabs:
> @chirag-silabs Fundamentally, What is causing the issue? DTIM sleep should be available during commissioning. We would like to get an explanation on why it is not possible with our wifi platform.

Actually, we have noticed that mdns discovery/CASE session happens a bit delayed or get missed. During the debugging of allegion we have seen that MDNS timeout/CASE session happens in high traffic environment like the ones in the QA labs. That's why it was suggested to be in high power till commissioning is done since it is a one time process. We are still debugging the wifi team about this but looks to be related to Network Congestion which will be debugged futhur.

I understand that it can still happen with the command noticed it happening where the mdns discovery happens after 14-16 sec and chip-tool shuts down at 20sec mark causing failure in hte CASE session. This is just to make sure that the device is always commissioned successfully at the first try itself since it is just a one time process(plus less time for commissioning)

jmartinez-silabs:
@chirag-silabs this pr brings a Build failure with GN for the 917 soc. See CI.
the 917 build step was working on previous commit. Can you fix that.
Only 917ncp was already broken

jmartinez-silabs (approved):

PR #215, opened 2025-01-20
[SL-UP] Moving the high-performance request command out of the timer handler

senthilku:
Description:
Issuing the power-save enable commands from the timer handler was causing a timeout issue, and the high-performance request before joining the network was failing.

Fix:
Moved the high-performance request command out of the timer handler.

Testing:
Tested commissioning and power cycling locally with the 917 SoC.

### DIFF
--- a/src/platform/silabs/wifi/SiWx/WifiInterface.cpp
+++ b/src/platform/silabs/wifi/SiWx/WifiInterface.cpp
@@ -462,12 +462,6 @@ sl_status_t JoinWifiNetwork(void)
 
     if (status == SL_STATUS_OK)
     {
-#if CHIP_CONFIG_ENABLE_ICD_SERVER
-        // TODO: We need a way to identify if this was a retry or a first attempt connect to avoid removing a req that was not ours
-        //  Remove High performance request that might have been added during the retry process
-        chip::DeviceLayer::Silabs::WifiSleepManager::GetInstance().RemoveHighPerformanceRequest();
-#endif // CHIP_CONFIG_ENABLE_ICD_SERVER
-
         WifiEvent event = WifiEvent::kStationConnect;
         sl_matter_wifi_post_event(event);
         return status;
--- a/src/platform/silabs/wifi/SiWx/WifiInterface.cpp
+++ b/src/platform/silabs/wifi/SiWx/WifiInterface.cpp
@@ -474,13 +474,6 @@ sl_status_t JoinWifiNetwork(void)
     // failure only happens when the firmware returns an error
     ChipLogError(DeviceLayer, "sl_net_up failed: 0x%lx", static_cast<uint32_t>(status));
 
-    // Deactivate the network interface before activating it on the next retry.
-    if ((status == SL_STATUS_SI91X_SCAN_ISSUED_IN_ASSOCIATED_STATE) || (status == SL_STATUS_SI91X_COMMAND_GIVEN_IN_INVALID_STATE))
-    {
-        status = sl_net_down((sl_net_interface_t) SL_NET_WIFI_CLIENT_INTERFACE);
-        ChipLogProgress(DeviceLayer, "sl_net_down status 0x%lx", static_cast<uint32_t>(status));
-    }
-
     wfx_rsi.dev_state.Clear(WifiState::kStationConnecting).Clear(WifiState::kStationConnected);
 
     ChipLogProgress(DeviceLayer, "Connection retry attempt %d", wfx_rsi.join_retries);
--- a/src/platform/silabs/wifi/SiWx/WifiInterface.cpp
+++ b/src/platform/silabs/wifi/SiWx/WifiInterface.cpp
@@ -819,9 +819,9 @@ void ProcessEvent(WifiEvent event)
     case WifiEvent::kStationStartJoin:
         ChipLogDetail(DeviceLayer, "WifiEvent::kStationStartJoin");
 
+// To avoid IOP issues, it is recommended to enable high-performance mode before joining the network.
+// TODO: Remove this once the IOP issue related to power save mode switching is fixed in the Wi-Fi SDK.
 #if CHIP_CONFIG_ENABLE_ICD_SERVER
-        // Request high-performance mode before joining the network
-        // Issuing power-save enable commands from the timer handler was causing a timeout issue; hence, it is moved here.
         chip::DeviceLayer::Silabs::WifiSleepManager::GetInstance().RequestHighPerformance();
 #endif // CHIP_CONFIG_ENABLE_ICD_SERVER
         InitiateScan();
--- a/src/platform/silabs/wifi/SiWx/WifiInterface.cpp
+++ b/src/platform/silabs/wifi/SiWx/WifiInterface.cpp
@@ -462,6 +462,10 @@ sl_status_t JoinWifiNetwork(void)
 
     if (status == SL_STATUS_OK)
     {
+#if CHIP_CONFIG_ENABLE_ICD_SERVER
+        // Remove High performance request that might have been added during the connect/retry process
+        chip::DeviceLayer::Silabs::WifiSleepManager::GetInstance().RemoveHighPerformanceRequest();
+#endif // CHIP_CONFIG_ENABLE_ICD_SERVER
         WifiEvent event = WifiEvent::kStationConnect;
         sl_matter_wifi_post_event(event);
         return status;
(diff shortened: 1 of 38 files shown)


senthilku (commented):
Observing some issue when testing with latest commit.

mkardous-silabs:
The PR changes the behavior of ICD exampe application but testing section does not define any tests that verify that sleep behavior is not significantly impacted.

- Does selective listening still work?
- Does the app go to sleep when it is expected to do so? what were the test scenarios?
- This change seem to conflict / duplicate the change chirag made, can we validate that both are requried?

mkardous-silabs (changes requested):
Requesting change until the validated scenarios are clarified and tested.

senthilku:
Does selective listening still work?

- Tested and selective listening feature is working fine.

Does the app go to sleep when it is expected to do so? what were the test scenarios?

 **The below scenarios are tested with lock application**

          - Device is going to sleep before commissioning
	  - Device is in Active stathe during commissioning
	  - Device is going to sleep (DTIM based ) after commissioning complete
	  - Device is going to sleep after power cycle
	  - Device is switching between Active and sleep mode during retry connection (active before connection and sleep after connection failure)
	  - Device is going to sleep after connection successful
	  - Device is going to selective LI with subscription enabled
	
This change seem to conflict / duplicate the change chirag made, can we validate that both are required?
  - yes, This change is required, Need to remove High performance request that might have been added during the connect/retry process, or else device will be always active after successful connection
  
What happens if we fail to join the Wi-Fi network, do we stay in high performance indefinitely?
 - Device is switching between Active and sleep mode during retry connection (active before connection and sleep after connection failure)

chirag-silabs (approved):

senthilku:
> The PR changes the behavior of ICD exampe application but testing section does not define any tests that verify that sleep behavior is not significantly impacted.
> 
> * Does selective listening still work?
Tested selective listening feature and it is working fine.
> * Does the app go to sleep when it is expected to do so? what were the test scenarios?
The below scenarios are tested with lock application
```   
  - Device is going to sleep before commissioning
  - Device is in Active stathe during commissioning
  - Device is going to sleep (DTIM based ) after commissioning complete
  - Device is going to sleep after power cycle
  - Device is switching between Active and sleep mode during retry connection (active before connection and sleep after connection failure)
  - Device is going to sleep after connection successful
  - Device is going to selective LI with subscription enabled
```
> * This change seem to conflict / duplicate the change chirag made, can we validate that both are requried?
yes, This change is required, Need to remove High performance request that might have been added during the connect/retry process, or else device will be always active after successful connection

> * What happens if we fail to join the Wi-Fi network, do we stay in high performance indefinitely?
Device is switching between Active and sleep mode during retry connection (active before connection and sleep after connection failure)

mkardous-silabs:
If i am not mistaken, this prevents the device from going to sleep while connecting to the Wi-Fi network. In practice, this shouldn't be the case. Can we move these high performance calls else where?

When we remove chirag's changes, this change will still impact power consumption and code execution.

senthilku:
> If i am not mistaken, this prevents the device from going to sleep while connecting to the Wi-Fi network. In practice, this shouldn't be the case. Can we move these high performance calls else where?
> 
> When we remove chirag's changes, this change will still impact power consumption and code execution.

It is expected to remain in a high-performance state while connecting to the Wi-Fi network.
In the existing code, this was achieved by calling the high-performance function from the timer handler during a retry connection. In the PR, the high-performance call was moved from the timer event handler because it is a blocking call, causing the high-performance request to fail (timeout) every time

mkardous-silabs:
I understand why the call cannot be in the timer callback; it shouldn't have been put there from the start.
I'm not sure i understand whyIt is expected to remain in a high-performance state while connecting to the Wi-Fi network.

Outisde of the Wi-Fi issue which Chirag opened a PR for, it is possible for the device to connect to Wi-Fi network without being in high-performance.

The goal of having the high performance req in the timer callback, even if it is wrong, was to take the device out of the sleep with no Wi-Fi network state. In other words, it aimed to solve a specific use-case. I don't think it it is necessary for the board to be in high performance. The change was simply "copied" from the previous logic to enable the board to sleep from the start.

The change you are proposing, changes the device connect behavior outside of just fixing the bug. I do not believe this is the correct way to fix this issue.

rosahay-silabs (approved):

senthilku:
> I understand why the call cannot be in the timer callback; it shouldn't have been put there from the start. I'm not sure i understand whyIt is expected to remain in a high-performance state while connecting to the Wi-Fi network.
> Wi-Fi team recomends to keep the High Performance mode before connecting to AP - https://jira.silabs.com/browse/SI91X-15845
> Outisde of the Wi-Fi issue which Chirag opened a PR for, it is possible for the device to connect to Wi-Fi network without being in high-performance.
> 
> The goal of having the high performance req in the timer callback, even if it is wrong, was to take the device out of the sleep with no Wi-Fi network state. In other words, it aimed to solve a specific use-case. I don't think it it is necessary for the board to be in high performance. The change was simply "copied" from the previous logic to enable the board to sleep from the start.
> 
> The change you are proposing, changes the device connect behavior outside of just fixing the bug. I do not believe this is the correct way to fix this issue.


-  Wi-Fi team recommends to keep the High-Performance mode before connecting to AP - https://jira.silabs.com/browse/SI91X-15845
- Switching between power modes in LMAC results in interoperability issues. Therefore, it is recommended to switch back to High Performance mode before connecting to the AP
- So, I have moved the High-Performance request to occur before connecting to avoid connection issues.

mkardous-silabs:
From the Jira ticket posted, 

>Had discussion with LMAC and Network Stack team regarding above observations. Switching from deep sleep to connected sleep in LMAC results in interop issues. It is recommended to switch back to High Performance mode before connecting to AP.

Deep Sleep is only during the retry attemps not during `happy path` were we are trying to connect for the first time.
Is the expectation to never go from sleep to connected?

senthilku:
> From the Jira ticket posted,
> 
> > Had discussion with LMAC and Network Stack team regarding above observations. Switching from deep sleep to connected sleep in LMAC results in interop issues. It is recommended to switch back to High Performance mode before connecting to AP.
> 
> Deep Sleep is only during the retry attemps not during `happy path` were we are trying to connect for the first time. Is the expectation to never go from sleep to connected?
> 
> Also, the ticket is for 3.3.X releases, is this still a limitation with the 3.4.X wifi sdk?

yes, we are observing this issue randomly when transitioning from sleep mode to connected mode.
It is an IOP issue, which is also mentioned in the Wi-Fi release notes here - https://confluence.silabs.com/display/EN/3.4.1+Release+Notes+-+SoC
image

src/platform/silabs/wifi/SiWx/WifiInterface.cpp
```diff
@@ -822,6 +826,11 @@ void ProcessEvent(WifiEvent event)
     case WifiEvent::kStationStartJoin:
         ChipLogDetail(DeviceLayer, "WifiEvent::kStationStartJoin");
 
+#if CHIP_CONFIG_ENABLE_ICD_SERVER
+        // Request high-performance mode before joining the network
```
mkardous-silabs, line 830: We should add an explanation why this is necessary.

src/platform/silabs/wifi/SiWx/WifiInterface.cpp
```diff
@@ -822,6 +826,11 @@ void ProcessEvent(WifiEvent event)
...
+        // Issuing power-save enable commands from the timer handler was causing a timeout issue; hence, it is moved here.
```
mkardous-silabs, line 831: This comment isn't necessary in the code.

src/platform/silabs/wifi/SiWx/WifiInterface.cpp
```diff
@@ -462,6 +462,10 @@ sl_status_t JoinWifiNetwork(void)
 
     if (status == SL_STATUS_OK)
     {
+#if CHIP_CONFIG_ENABLE_ICD_SERVER
+        // Remove High performance request that might have been added during the connect/retry process
+        chip::DeviceLayer::Silabs::WifiSleepManager::GetInstance().RemoveHighPerformanceRequest();
+#endif // CHIP_CONFIG_ENABLE_ICD_SERVER
```
mkardous-silabs, line 468: I'm not sure i understand why we remove a high performance request here.
We should only remove when we have connected to the wifi network.

I don't think this should here.

senthilku: We enabled the high-performance request before the join call and are removing it here if the connection is successful.

mkardous-silabs (approved):

PR #227, opened 2025-01-21
[SL-UP] Dimmer switch button implementation

arun-silabs:
Description : This PR extends the button implementation in dimmer switch app by adding level control actions on button sequences. Below are the implementation details : 

Default Step direction - Up
Button1 short press - light toggle
Button1 long press - Level up/down as per direction
Button1 shortpress + Button0 shortpress - reverse the direction (down to up or up to down)
Button0 longpress - Factory reset

Testing : Tested the button sequences on SOC manually.

### DIFF
--- a/examples/light-switch-app/silabs/include/LightSwitchMgr.h
+++ b/examples/light-switch-app/silabs/include/LightSwitchMgr.h
@@ -84,6 +84,8 @@ public:
     void TriggerLightSwitchAction(LightSwitchAction action, bool isGroupCommand = false);
     void TriggerLevelControlAction(StepModeEnum stepMode, bool isGroupCommand = false);
 
+    StepModeEnum getStepMode();
+
     AppEvent CreateNewEvent(AppEvent::AppEventTypes type);
 
     static LightSwitchMgr & GetInstance() { return sSwitch; }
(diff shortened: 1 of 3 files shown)


examples/light-switch-app/silabs/include/LightSwitchMgr.h
```diff
@@ -103,9 +105,12 @@ class LightSwitchMgr
 private:
     static LightSwitchMgr sSwitch;
 
-    Timer * mLongPressTimer = nullptr;
-    bool mDownPressed       = false;
-    bool mResetWarning      = false;
+    Timer * mLongPressTimer    = nullptr;
+    bool mDownPressed          = false;
```
jmartinez-silabs, line 109: Add a comment for `mDownPressed  mUpPressed mUpSuppressed` are/mean. Also I think your took those names from the window app. but here a feel they aren't the right name for what they are. 
Down Up are not relevant here

this applies to all naming referencing Up vs Down

jmartinez-silabs (changes requested):

examples/light-switch-app/silabs/include/LightSwitchMgr.h
arun-silabs: Modified the names to imply the levelup and leveldown actions

jmartinez-silabs (approved):

PR #238, opened 2025-01-24
[SL-ONLY] Add workflow to open PR from csa to main after the csa update is completed

mkardous-silabs:
### Description
PR adds a workflow that is triggered after the daily csa update to open a PR from csa to main with the latest CSA changes.

When testing the action I saw an issue with the CI was not being triggered on the PR opened by a Github Action but i think having the PAT token checkout and create the branch fixes the issue. With all the testing i can confirm a 100% this.
We will know tomorrow morning if the full flow works.

### Tests
Check https://github.com/SiliconLabsSoftware/matter_sdk/pull/236

### DIFF
--- a/.github/workflows/silabs-open-csa-pr.yaml
+++ b/.github/workflows/silabs-open-csa-pr.yaml
@@ -39,3 +39,16 @@ jobs:
                   token: ${{secrets.GITHUB_TOKEN}}
                   branch-token: ${{secrets.WORKFLOW_TOKEN}}
                   labels: changing-submodules-on-purpose
+
+            - name: Comment on Pull Request
+              uses: actions/github-script@v6
+              with:
+                  github-token: ${{secrets.GITHUB_TOKEN}}
+                  script: |
+                      const pr_number = context.payload.workflow_run.pull_requests[0].number;
+                      github.issues.createComment({
+                        owner: context.repo.owner,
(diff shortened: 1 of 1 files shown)


lpbeliveau-silabs (approved):

.github/workflows/silabs-open-csa-pr.yaml
```diff
@@ -0,0 +1,41 @@
...
+                  branch: automation/update_main
+                  base: main
+                  title: "Sync csa branch with main"
+                  body: |
+                      This PR syncs the csa branch with the main branch.
+
+                      **PR MUST BE MERGED WITH MERGE COMMIT - ADMIN MUST ENABLE THE OPTION**
+                  token: ${{secrets.GITHUB_TOKEN}}
+                  branch-token: ${{secrets.WORKFLOW_TOKEN}}
+                  labels: changing-submodules-on-purpose
```
jmartinez-silabs, line 41: Can we add a label with maybe a special CI check that require admin merge for to ensure standard merge rather than squash merge

jmartinez-silabs (approved):

.github/workflows/silabs-open-csa-pr.yaml
mkardous-silabs: added step to comment PR to remind merge commit

PR #254, opened 2025-01-27
[SL-UP] Raise TimerTask to the highest prio

jmartinez-silabs:
Raise the timer task to the highest priority. 

some slcc/slcp try to set configTIMER_TASK_PRIORITY for some stacks.
For example 
```
Timer RTOS task priority needs to be one more than SL_BT_RTOS_LINK_LAYER_TASK_PRIORITY
BT uses event flags that require the timer task to be a higher priority than all other RTOS tasks
```
In matter extension, we overwrite the generated freeRTOSConfig but our static config. This PR makes sure this prio is always set.

#### Testing
Manual commissioning and control tests.

### DIFF
--- a/examples/platform/silabs/FreeRTOSConfig.h
+++ b/examples/platform/silabs/FreeRTOSConfig.h
@@ -174,11 +174,8 @@ extern uint32_t SystemCoreClock;
 
 /* Software timer related definitions. */
 #define configUSE_TIMERS (1)
-#ifdef SLI_SI917
+// Keep the timerTask at the highest prio as some of our stacks tasks leverage eventing with timers.
 #define configTIMER_TASK_PRIORITY (55) /* Highest priority */
-#else
-#define configTIMER_TASK_PRIORITY (40) /* Highest priority */
-#endif                                 // SLI_SI917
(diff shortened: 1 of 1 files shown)


mkardous-silabs (approved):

examples/platform/silabs/FreeRTOSConfig.h
```diff
@@ -313,7 +310,7 @@ standard names. */
 /* Thread local storage pointers used by the SDK */
 
 #ifndef configNUM_USER_THREAD_LOCAL_STORAGE_POINTERS
-#define configNUM_USER_THREAD_LOCAL_STORAGE_POINTERS 2
+#define configNUM_USER_THREAD_LOCAL_STORAGE_POINTERS 0
```
lpbeliveau-silabs, line 313: I take it that we didn't need any local storage pointers for thread?

jmartinez-silabs: From what I can tell no. we don't use it. The slc generated one doesn't set any. I'll be slowly trying to clean up this config and see if we can use the generated one directly

PR #258, opened 2025-01-28
[SL-ONLY] Tracing in CI

lpbeliveau-silabs:
Fixed tracing build and add tracing build option to CI to prevent future breakage

#### Testing

Built efr32 NCP apps, efr32 mg24 appps and efr32 mg26 apps with and without tracing enabled.

CI will tell if I enabled it properly.

### DIFF
--- a/.github/silabs-builds-mg26.json
+++ b/.github/silabs-builds-mg26.json
@@ -4,5 +4,11 @@
             "boards": ["BRD4116A", "BRD4117A", "BRD4118A", "BRD2608A"],
             "arguments": ["--docker"]
         }
+    ],
+    "lighting-app":[
+        {
+            "boards": ["BRD2608A"],
+            "arguments": ["--docker", "chip_build_libshell=true", "matter_enable_tracing_support=true", "--clean"]
+        }
     ]
 }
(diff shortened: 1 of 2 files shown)


scripts/examples/gn_silabs_example.sh
mkardous-silabs, line 1: If we are not going to upstream the tracing changes. It might be worth while not to change this file and just use the gn arguments directly in the CI and manually.

.github/silabs-builds-mg26.json
```diff
@@ -3,6 +3,10 @@
         {
             "boards": ["BRD4116A", "BRD4117A", "BRD4118A", "BRD2608A"],
             "arguments": ["--docker"]
+        },
+        {
+            "boards": ["BRD2608A"],
+            "arguments": ["--docker", "--enable_tracing", "--clean"]
         }
```
mkardous-silabs, line 10: This is going to run for all samples. Should we run it only for the ligthing app instead?

mkardous-silabs (approved):

jmartinez-silabs (approved):

PR #259, opened 2025-01-28
[SL-ONLY] Add require admin check workflow

mkardous-silabs:
### Description
PR adds a workflow that checks if the `sl-require-admin-action` label is present. If the label is present, the workflow fails to force admin action.

PR adds two jobs, one to check if the label is present and a second that checks if the label was removed.

The goal of this action is to force an admin validation on certain situations like the update of main from csa that requires a special merge strategy.


.github/workflows/silabs-require-admin-action-check.yaml
```diff
@@ -0,0 +1,58 @@
...
+on:
+    pull_request:
+        branches:
+            - main
+            - release_*
+        types:
+            - opened
+            - reopened
+            - synchronize
+            - labeled
```
jmartinez-silabs, line 12: nice

jmartinez-silabs (approved):

lpbeliveau-silabs (approved):

PR #268, opened 2025-02-04
[SL-UP][WiFi] Fixing the rssi value for wf200

chirag-silabs:
WF200 scan was sending RCPI value rather than the RSSI value. Fixing it so that the send value is RSSI

Formula for the RSSI
```
The following equation can be used to convert the RCPI value to corresponding dBm value.
[signal level in dBm] = (RCPI / 2) - 110
```
https://docs.silabs.com/wifi/wf200/rtos/latest/group-w-f-m-g-r-o-u-p-c-o-n-c-e-p-t-s

Tested on WF200

### DIFF
--- a/src/platform/silabs/wifi/WifiInterfaceAbstraction.h
+++ b/src/platform/silabs/wifi/WifiInterfaceAbstraction.h
@@ -301,12 +301,6 @@ void sl_button_on_change(uint8_t btn, uint8_t btnAction);
 #ifdef WF200_WIFI
 void sl_wfx_host_gpio_init(void);
 void wfx_bus_start(void);
-/**
- * @brief Convert RCPI to RSSI
- * @param[in]  rcpi: Received Channel Power Indicator value
- * @return RSSI value
- */
-inline int32_t ConvertRcpiToRssi(int32_t rcpi) { return (rcpi / 2) - 110; }
 #endif /* WF200_WIFI */
 
 #ifdef __cplusplus
--- a/src/platform/silabs/wifi/WifiInterfaceAbstraction.h
+++ b/src/platform/silabs/wifi/WifiInterfaceAbstraction.h
@@ -301,6 +301,12 @@ void sl_button_on_change(uint8_t btn, uint8_t btnAction);
 #ifdef WF200_WIFI
 void sl_wfx_host_gpio_init(void);
 void wfx_bus_start(void);
+/**
+ * @brief Convert RCPI to RSSI
+ * @param[in]  rcpi: Received Channel Power Indicator value
+ * @return RSSI value
+ */
+inline int32_t ConvertRcpiToRssi(int32_t rcpi) { return (rcpi / 2) - 110; }
 #endif /* WF200_WIFI */
 
 #ifdef __cplusplus
--- a/src/platform/silabs/wifi/wf200/WifiInterface.cpp
+++ b/src/platform/silabs/wifi/wf200/WifiInterface.cpp
@@ -479,7 +479,7 @@ static void sl_wfx_scan_result_callback(sl_wfx_scan_result_ind_body_t * scan_res
             ap->scan.security = WFX_SEC_NONE;
         }
         ap->scan.chan = scan_result->channel;
-        ap->scan.rssi = scan_result->rcpi;
+        ap->scan.rssi = (scan_result->rcpi)/2 - 110;
         memcpy(&ap->scan.bssid[0], &scan_result->mac[0], BSSID_LEN);
         scan_count++;
     }
--- a/src/platform/silabs/wifi/wf200/WifiInterface.cpp
+++ b/src/platform/silabs/wifi/wf200/WifiInterface.cpp
@@ -913,7 +913,7 @@ int32_t wfx_get_ap_info(wfx_wifi_scan_result_t * ap)
     sl_status_t status = sl_wfx_get_signal_strength((uint32_t *) &signal_strength);
     VerifyOrReturnError(status == SL_STATUS_OK, status);
     ChipLogDetail(DeviceLayer, "signal_strength: %ld", signal_strength);
-    ap->rssi = (signal_strength - 220) / 2;
+    ap->rssi = (signal_strength / 2) - 110;
     return status;
 }
 
--- a/src/platform/silabs/wifi/wf200/WifiInterface.cpp
+++ b/src/platform/silabs/wifi/wf200/WifiInterface.cpp
@@ -300,7 +300,7 @@ error_handler:
  */
 inline int16_t ConvertRcpiToRssi(uint32_t rcpi) {
     int64_t rssi = (rcpi / 2) - 110;
-    // Checking for the overflows
+    // Checking for overflows
     VerifyOrReturnValue(rssi < std::numeric_limits<int16_t>::max(), std::numeric_limits<int16_t>::max());
     VerifyOrReturnValue(rssi > std::numeric_limits<int16_t>::min(), std::numeric_limits<int16_t>::min());
     return rssi;
--- a/src/platform/silabs/wifi/wf200/WifiInterface.cpp
+++ b/src/platform/silabs/wifi/wf200/WifiInterface.cpp
@@ -292,6 +292,15 @@ error_handler:
 
     return result;
 }
+
+/**
+ * @brief Convert RCPI to RSSI
+ * @param[in]  rcpi: Received Channel Power Indicator value
+ * @return RSSI value
+ */
+int32_t ConvertRcpiToRssi(int32_t rcpi) {
+    return (rcpi / 2) - 110;
+}
 } // namespace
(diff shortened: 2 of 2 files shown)


mkardous-silabs (changes requested):
RSSI is calculated in two different locations. The fix should be donne in both location.

chirag-silabs:
> RSSI is calculated in two different locations. The fix should be donne in both location.

The calution was correct at the other place, anyhow made it similar to match the formula

mkardous-silabs:
@chirag-silabs I would suggest having an inline function that does the calcluation instead of implementing the calculation twice.

function takes the rcpi and returns rssi.

src/platform/silabs/wifi/WifiInterfaceAbstraction.h
```diff
@@ -301,6 +301,12 @@ void sl_button_on_change(uint8_t btn, uint8_t btnAction);
 #ifdef WF200_WIFI
 void sl_wfx_host_gpio_init(void);
 void wfx_bus_start(void);
+/**
+ * @brief Convert RCPI to RSSI
+ * @param[in]  rcpi: Received Channel Power Indicator value
+ * @return RSSI value
+ */
+inline int32_t ConvertRcpiToRssi(int32_t rcpi) { return (rcpi / 2) - 110; }
```
mkardous-silabs, line 309: We don't need the signature to be in the abstraction header since it is only used in the wf200 interface.
This should on the WifiInterface files in the anounymous space of the .cpp.

src/platform/silabs/wifi/wf200/WifiInterface.cpp
```diff
@@ -293,6 +293,14 @@ sl_status_t get_all_counters(void)
     return result;
 }
 
+/**
+ * @brief Convert RCPI to RSSI
+ * @param[in]  rcpi: Received Channel Power Indicator value
+ * @return RSSI value
+ */
+int32_t ConvertRcpiToRssi(int32_t rcpi) {
```
mkardous-silabs, line 301: Incorrect type for the rcpi input.

chirag-silabs: corrected the type

jmartinez-silabs (approved):

arun-silabs (approved):

src/platform/silabs/wifi/wf200/WifiInterface.cpp
```diff
@@ -436,7 +444,7 @@ static void sl_wfx_scan_result_callback(sl_wfx_scan_result_ind_body_t * scan_res
     struct scan_result_holder * ap;
 
     ChipLogDetail(DeviceLayer, "# %2d %2d  %03d %02X:%02X:%02X:%02X:%02X:%02X  %s", scan_count, scan_result->channel,
-                  ((int16_t) (scan_result->rcpi - 220) / 2), scan_result->mac[0], scan_result->mac[1], scan_result->mac[2],
+                  ((int16_t) (ConvertRcpiToRssi(scan_result->rcpi))), scan_result->mac[0], scan_result->mac[1], scan_result->mac[2],
```
mkardous-silabs, line 447: Is there int16_t still necessary here since the function returns a int16_t?

chirag-silabs: removed.

src/platform/silabs/wifi/wf200/WifiInterface.cpp
```diff
@@ -896,7 +904,7 @@ static void wfx_wifi_hw_start(void)
  **************************************************************************/
 int32_t wfx_get_ap_info(wfx_wifi_scan_result_t * ap)
 {
-    int32_t signal_strength;
+    uint16_t signal_strength;
```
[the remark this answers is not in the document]
chirag-silabs: updated to uint32_t, noticed in the scan callback it was mentioned to be uint16_t so kept it to be uint16_t

src/platform/silabs/wifi/wf200/WifiInterface.cpp
```diff
@@ -912,8 +920,8 @@ int32_t wfx_get_ap_info(wfx_wifi_scan_result_t * ap)
 
     sl_status_t status = sl_wfx_get_signal_strength((uint32_t *) &signal_strength);
```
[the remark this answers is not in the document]
chirag-silabs: Done

src/platform/silabs/wifi/wf200/WifiInterface.cpp
```diff
@@ -293,6 +293,17 @@ sl_status_t get_all_counters(void)
     return result;
 }
 
+/**
+ * @brief Convert RCPI to RSSI
+ * @param[in]  rcpi: Received Channel Power Indicator value
+ * @return RSSI value
+ */
+int16_t ConvertRcpiToRssi(uint32_t rcpi) {
```
mkardous-silabs, line 301: Function isn't inline anymore, what was that intentionnal?

src/platform/silabs/wifi/wf200/WifiInterface.cpp
```diff
@@ -293,6 +293,17 @@ sl_status_t get_all_counters(void)
...
+    // Checking for the overflows
+    // If rcpi is greater than or equal to INT16_MAX, return INT16_MIN to indicate an error
+    VerifyOrReturnValue(rcpi < INT16_MAX, INT16_MIN);
```
mkardous-silabs, line 304: For reference, `std::numeric_limits` is the C++ way to get the type max and min values.

src/platform/silabs/wifi/wf200/WifiInterface.cpp
```diff
@@ -293,6 +293,17 @@ sl_status_t get_all_counters(void)
...
 }
 
+/**
+ * @brief Convert RCPI to RSSI
+ * @param[in]  rcpi: Received Channel Power Indicator value
+ * @return RSSI value
+ */
+int16_t ConvertRcpiToRssi(uint32_t rcpi) {
+    // Checking for the overflows
+    // If rcpi is greater than or equal to INT16_MAX, return INT16_MIN to indicate an error
```
mkardous-silabs, line 303: Even if in theory this should never happen, we are not interested of checking if the input if bigger than a int16. We need to check that the output is.

In others, do the math with an int32 (need to check if 64 isn't better here), if the result is bigguer than a int16 return an error.

src/platform/silabs/wifi/wf200/WifiInterface.cpp
mkardous-silabs, line 301: If we want to return errors, we shouldn't return an arbitrary value. The signature should be :
```
CHIP_ERROR ConvertRCPIToRssi(uint32_t rcpi, int16_t & rssi);
```

Then again, since this is a local function, it is good enough to just set the closest value to reduce the complexity.
In other words, if the result is < int16_min, set int16_min. If the result is >int16_max, set int16_max.

src/platform/silabs/wifi/wf200/WifiInterface.cpp
chirag-silabs: added std::numeric_limits

src/platform/silabs/wifi/wf200/WifiInterface.cpp
chirag-silabs: added limit checks for the return

src/platform/silabs/wifi/wf200/WifiInterface.cpp
chirag-silabs: done

src/platform/silabs/wifi/wf200/WifiInterface.cpp
```diff
@@ -293,6 +293,18 @@ sl_status_t get_all_counters(void)
...
 }
 
+/**
+ * @brief Convert RCPI to RSSI
+ * @param[in]  rcpi: Received Channel Power Indicator value
+ * @return RSSI value
+ */
+inline int16_t ConvertRcpiToRssi(uint32_t rcpi) {
+    int64_t rssi = (rcpi / 2) - 110;
+    // Checking for the overflows
```
mkardous-silabs, line 303:
```suggestion
    // Checking for overflows
```

src/platform/silabs/wifi/wf200/WifiInterface.cpp
```diff
@@ -293,6 +293,18 @@ sl_status_t get_all_counters(void)
     return result;
 }
 
+/**
+ * @brief Convert RCPI to RSSI
```
mkardous-silabs, line 297: Missing comments when the result of the formula doesn't fit in a int16

chirag-silabs: Added comments

mkardous-silabs (approved):

PR #281, opened 2025-02-13
[SL-UP] Added a function that can empty the uart in a blocking manner for ins…

lpbeliveau-silabs:
#### Testing

Tested by placing buffer flush before reboot and removing the os delays, the reboot will be delayed until the logging is done.

Was tested with Silabs Tracing for OTA as well but this will be in an SL-TEMP PR.

### DIFF
--- a/examples/platform/silabs/uart.cpp
+++ b/examples/platform/silabs/uart.cpp
@@ -583,6 +583,19 @@ void uartSendBytes(uint8_t * buffer, uint16_t nbOfBytes)
 #endif // SLI_SI91X_MCU_INTERFACE
 }
 
+/**
+ * @brief Flush the UART TX queue in a blocking manner.
+ */
+void uartFlushTxQueue(void)
+{
+    UartTxStruct_t workBuffer;
+
+    while (osMessageQueueGet(sUartTxQueue, &workBuffer, nullptr, 0) == osOK)
+    {
+        UARTDRV_ForceTransmit(vcom_handle, workBuffer.data, workBuffer.length);
+    }
+}
+
 #ifdef __cplusplus
 }
 #endif
--- a/examples/platform/silabs/uart.h
+++ b/examples/platform/silabs/uart.h
@@ -28,6 +28,7 @@ void uartConsoleInit(void);
 int16_t uartConsoleWrite(const char * Buf, uint16_t BufLength);
 int16_t uartLogWrite(const char * log, uint16_t length);
 int16_t uartConsoleRead(char * Buf, uint16_t NbBytesToRead);
+void uartFlushTxQueue(void);
 
 void uartMainLoop(void * args);
 


jmartinez-silabs (approved):

examples/platform/silabs/uart.cpp
```diff
@@ -195,7 +199,7 @@ static Fifo_t sReceiveFifo;
 #if SLI_SI91X_MCU_INTERFACE == 0
 static void UART_rx_callback(UARTDRV_Handle_t handle, Ecode_t transferStatus, uint8_t * data, UARTDRV_Count_t transferCount);
 #endif // SLI_SI91X_MCU_INTERFACE == 0
-static void uartSendBytes(uint8_t * buffer, uint16_t nbOfBytes);
+static void uartSendBytes(UartTxStruct_t * bufferStruct, uint16_t nbOfBytes);
```
jmartinez-silabs, line 202: since UartTxStruct_t has the length already we could drop `nbOfBytes` param

examples/platform/silabs/uart.cpp
```diff
@@ -533,7 +537,7 @@ void uartMainLoop(void * args)
         osStatus_t eventReceived = osMessageQueueGet(sUartTxQueue, &workBuffer, nullptr, osWaitForever);
         while (eventReceived == osOK)
         {
-            uartSendBytes(workBuffer.data, workBuffer.length);
+            uartSendBytes(&workBuffer, workBuffer.length);
```
jmartinez-silabs, line 540: it would become uartSendBytes(&workBuffer);

examples/platform/silabs/uart.cpp
```diff
@@ -542,18 +546,18 @@ void uartMainLoop(void * args)
...
+ * @param buffer pointer to the UartTxStruct_t containing the data
  * @param nbOfBytes number of bytes to send
  */
-void uartSendBytes(uint8_t * buffer, uint16_t nbOfBytes)
+void uartSendBytes(UartTxStruct_t * bufferStruct, uint16_t nbOfBytes)
 {
 #if SLI_SI91X_MCU_INTERFACE
     // ensuring null termination of buffer
-    if (nbOfBytes != CHIP_SHELL_MAX_LINE_SIZE && buffer[nbOfBytes - 1] != '\0')
+    if (nbOfBytes < sizeof(bufferStruct->data) && bufferStruct->data[nbOfBytes - 1] != '\0')
```
jmartinez-silabs, line 556: sizeof(bufferStruct->data) this would give you 4 as size of the pointer no ?
if for some reason not caught before, nbOfBytes is > than bufferStruct->data we aren't ensuring that bufferStruct->data is null terminated and we are still passing it to Board_UARTPutSTR. That will cause out of memory access because I believe Board_UARTPutSTR depends on the null char

jmartinez-silabs (changes requested):

examples/platform/silabs/uart.cpp
lpbeliveau-silabs: True, somehow I thought length was size when I edited that

examples/platform/silabs/uart.cpp
lpbeliveau-silabs: size data is a defined sized array : data[UART_TX_MAX_BUF_LEN], this will return the total size of the entire array.

examples/platform/silabs/uart.cpp
```diff
@@ -542,18 +545,21 @@ void uartMainLoop(void * args)
...
+    {
+        bufferStruct->data[bufferStruct->length] = '\0';
+    }
+    else
     {
-        buffer[nbOfBytes] = '\0';
+        bufferStruct->data[ArraySize(bufferStruct->data) - 1] = '\0';
     }
-    Board_UARTPutSTR(reinterpret_cast<uint8_t *>(buffer));
+    Board_UARTPutSTR(reinterpret_cast<uint8_t *>(bufferStruct->data));
```
jmartinez-silabs, line 562: isn't `bufferStruct->data` already a uint8_t* ?

jmartinez-silabs (approved):

examples/platform/silabs/uart.cpp
lpbeliveau-silabs: Yes it is

mkardous-silabs (approved):

rcasallas-silabs (approved):

examples/platform/silabs/uart.cpp
```diff
@@ -542,18 +545,21 @@ void uartMainLoop(void * args)
 /**
  * @brief Send Bytes to UART. This blocks the UART task.
  *
- * @param buffer pointer to the buffer containing the data
- * @param nbOfBytes number of bytes to send
+ * @param buffer pointer to the UartTxStruct_t containing the data
  */
-void uartSendBytes(uint8_t * buffer, uint16_t nbOfBytes)
+void uartSendBytes(UartTxStruct_t * bufferStruct)
```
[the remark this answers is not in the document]
jmartinez-silabs: Pretty sure we save some stack here by passing the pointer to the bufferStruct rather than copying it.

PR #287, opened 2025-02-17
[SL-TEMP] Blocking Trace Flush

lpbeliveau-silabs:
Made the Tracing Flush blocking when used with UART Logs to ensure the trace are logged before a reboot during an OTA.

Fixes: https://jira.silabs.com/browse/MATTER-4524

#### Testing

Tested the output of the lofs while performing and OTA from a concurrent CMP app.


examples/platform/silabs/uart.cpp
```diff
@@ -20,6 +20,7 @@
 #include "MatterShell.h" // nogncheck
 #endif
 #include <cmsis_os2.h>
+#include <lib/support/CodeUtils.h>
```
rcasallas-silabs, line 23: Is this change intended?

rcasallas-silabs (approved):

jmartinez-silabs (approved):

jmartinez-silabs:
Why is it SL_TEMP?

examples/platform/silabs/uart.cpp
lpbeliveau-silabs: Interesting... should have been in my previous PR, but yes. This is to use the ArraySize() maccro.
The fact that this builds without it means that something else is including it but I'd rather have the include in here.

PR #288, opened 2025-02-17
[SL-UP] Removing the IP polling from the Matter code

chirag-silabs:
The IP polling and assignment is already a part of the wifi SDK code, so removing the code from the matter layer.

Keeping the notify code which is used for the notify the application about the connectivity

#### Testing
Multiple smoke runs were done to validate the change.
Tested on the 917 SoC

### DIFF
--- a/src/platform/silabs/wifi/SiWx/WifiInterface.cpp
+++ b/src/platform/silabs/wifi/SiWx/WifiInterface.cpp
@@ -100,10 +100,6 @@ bool ps_requirement_added = false;
 #endif // CHIP_CONFIG_ENABLE_ICD_SERVER
 
 bool hasNotifiedWifiConnectivity = false;
-bool hasNotifiedIPV6             = false;
-#if (CHIP_DEVICE_CONFIG_ENABLE_IPV4)
-bool hasNotifiedIPV4 = false;
-#endif /* CHIP_DEVICE_CONFIG_ENABLE_IPV4 */
 
 wfx_wifi_scan_ext_t temp_reset;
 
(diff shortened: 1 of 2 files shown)


jmartinez-silabs:
This can be done on CSA now. We are also not in a release cycle so it would be preferable to do it on CSA first.

src/platform/silabs/wifi/SiWx/WifiInterface.cpp
```diff
@@ -724,14 +657,9 @@ void sl_matter_wifi_post_event(WifiEvent event)
 ///        and emits a WifiEvent::kStationDoDhcp event to trigger DHCP polling checks. Helper function for ProcessEvent.
 void ResetDHCPNotificationFlags(void)
```
rosahay-silabs, line 658: if DHCP related process is missing do we need this API? if yes, then it should be renamed.

src/platform/silabs/wifi/SiWx/WifiInterface.cpp
```diff
@@ -661,53 +630,17 @@ sl_status_t bg_scan_callback_handler(sl_wifi_event_t event, sl_wifi_scan_result_
 
 /// NotifyConnectivity
 /// @brief Notify the application about the connectivity status if it has not been notified yet.
-///        Helper function for HandleDHCPPolling.
 void NotifyConnectivity(void)
 {
     VerifyOrReturn(!hasNotifiedWifiConnectivity);
+#if (CHIP_DEVICE_CONFIG_ENABLE_IPV4)
```
rosahay-silabs, line 636: according to existing logic, we used to trigger these based on the relevant IPv4 and IPv6 events, this gives me the impression that no matter what both IPv4 IPv6 has been acquired, is that understanding correct?
If yes, then do we need to keep track of the `hasNotifiedWifiConnectivity` event at all?

rosahay-silabs (changes requested):

PR #305, opened 2025-02-26
[SL-ONLY] Fix build when logging is didisabled

chirag-silabs:
#### Testing

The build was failing when `SILABS_LOG_ENABLED` was defined to 0 

Correcting the condition to fix the build

### DIFF
--- a/src/platform/silabs/Logging.cpp
+++ b/src/platform/silabs/Logging.cpp
@@ -196,7 +196,7 @@ extern "C" void silabsLog(const char * aFormat, ...)
     va_end(v);
 }
 
-#ifdef SILABS_LOG_ENABLED
+#if SILABS_LOG_ENABLED
 bool isLogInitialized()
 {
     return sLogInitialized;


rosahay-silabs (approved):
fix the title of the PR

mkardous-silabs:
This isn't shouldn't go in main but in CSA. Bahvani has an sl-up todo to bring this to CSA.

mkardous-silabs (changes requested):
See comment

chirag-silabs:
> This isn't shouldn't go in main but in CSA. Bahvani has an sl-up todo to bring this to CSA.

The PR is for SL-ONLY itself. Discussed with Bhavani, by mistake she added the fix in same PR of SL-UP rather than splitting it
Here's the PR which added this
https://github.com/SiliconLabsSoftware/matter_sdk/commit/a70f40e3b498d537bfcfdd3002ea13e00d14a395
therefore it should be SL-ONLY commit

mkardous-silabs (approved):

PR #312, opened 2025-03-04
[SL-TEMP] Increasing the lwip thread priority to 32 (osPriorityAboveNormal)

chirag-silabs:
#### Testing

After the move to the cmsis os the lwip priority was still 2 i.e., below the application (osPriorityNormal)
Increasing the lwip task priority to 32

Current task priorities list is as follows:

Timer SVC - 55 (Highest)
si91x_event - 49 : Handles the notifications event coming from the wifi sdk itself
si91x_bus - 48 : Handles the Tx and Rx events coming from the wifi sdk itself
rsi_ble - 40 : Matter platform layer for the ble events, this internalls calls APIs from the wifi sdk and vice versa
wlan_rsi - 39 : Matter platform layer for the wlan events, this internalls calls APIs from the wifi sdk and vice versa
LIT - 24 : Application thread
Shell - 16 : Matter CLI
UART - 16 : Logging
LWIP - 2 : Lwip packet handling low level input and output handling
CHIP - 2 : Matter thread common for all the platforms. Handling the data model read and writes. Provided by chip stack itself.
IDLE - 0

After discussing with the wifi sdk team it was suggested to keep the priority more than the application, there can be possibility that the lwip is blocked by the application thread if throughtput is more and your application is processing multiple events simulateously.

I have increased the priority to 32 and have validated commissioning and commands are working fine.

Currently since we don't get many events on the application thread like btn and all which is not affecting it but it can affect in the future when the application is more events blocking lwip.

Testing
Tested with 917 SoC and Wf200 RCP

Multiple commissioning runs and commands transfer.

### Note:

This is a temporary change for Matter 1.4, the CSA code have chip task of priority osPriorityHigh so it should be analyzed and all the other tasks should also be modified.


src/lwip/silabs/lwipopts-wf200.h
```diff
@@ -69,18 +70,19 @@
...
 #define MEM_SIZE 5632
 #define MEMP_NUM_UDP_PCB (6)
 #define TCP_MSS (4 * 1152)
-#define TCPIP_THREAD_PRIO (3)
 #else
 #define LWIP_DNS 0
 #define MEMP_NUM_UDP_PCB (5)
 #define TCP_MSS (1152)
-#define TCPIP_THREAD_PRIO (2)
+
```
rosahay-silabs, line 85:
```suggestion
```

jmartinez-silabs (approved):

PR #319, opened 2025-03-07
[SL-TEMP] OT_NETIF_THREAD_HOST

lpbeliveau-silabs:
Updated `OT_NETIF_THREAD` define to match changes in OT stack update.

#### Testing
 Compile->Flash->Commission->Toggle


selissia (approved):

jmartinez-silabs:
I think we should do a sl-temp with this one. When we update csa with the new sisdk that needs this. The ifdef define won't be needed and the new define usage in the API will be needed to build so we know this won't be forgotten.

jmartinez-silabs (approved):

PR #333, opened 2025-03-11
[CSA-CP] temporary workaround: keeping pygobject==3.50.0 to avoid dependency i…

mkardous-silabs:
…ssues (#37948)

Update of the pygobject dependencies broke the CSA CI for python builds. This forces a specific version to fix the issue until we update to the latest version.


chirag-silabs (approved):

jmartinez-silabs:
I see a couple PR that bring the same thing for bord 2.5 and 2.6 branch. I also need to do a couple CSA CP for those 2 branch. I think we should just plan a merge for 2.5 in 2.6 by the end of the week and reduce 2 duplicate PR now for 2.6

PR #339, opened 2025-03-12
[SL-TEMP] Changes 917 SoC/NCP sleep config to connected sleep

bhmanda-silabs:
**Description**

- In the ConfigureDeepSleep function of the SiWx platform, setting to BLE sleep config to RSI_SLEEP_MODE_8 which is deep sleep when DUT is not provisioned. So this prevents the initial BLE connection to work since DUT can't advertise in this mode.
- This PR changes the default for all sleep configs to connected sleep which is RSI_SLEEP_MODE_2 till commissioning completed so that ble advertising and connection are working.
#### Testing
Verified commissioning on 917 SoC and NCP boards and commands.


mkardous-silabs:
@bhmanda-silabs I'm not sure i understand the way the issue is being fixed here. We can fix the issue by always configuring BLE in the associated sleep mode which simplifies our logic. 

This would also mean that all the ifdef in the ApplicationSleepManager can be removed.

src/platform/silabs/wifi/WifiInterfaceAbstraction.h
```diff
@@ -118,9 +118,9 @@ typedef enum
 
 typedef struct
 {
-    char ssid[WFX_MAX_SSID_LENGTH + 1];
+    char ssid[WFX_MAX_SSID_LENGTH + 1] = {0};
```
senthilku, line 121: Why all the elements of the ssid array is initialized to zero? observed any issue?

examples/platform/silabs/wifi/icd/ApplicationSleepManager.cpp
```diff
@@ -53,17 +53,17 @@ CHIP_ERROR ApplicationSleepManager::Init()
 void ApplicationSleepManager::OnCommissioningWindowOpened()
 {
     mIsCommissionningWindowOpen = true;
-#if (defined(SLI_SI91X_MCU_INTERFACE) && SLI_SI91X_MCU_INTERFACE == 1)
+#if (defined(SLI_SI917) && SLI_SI917 == 1)
```
senthilku, line 56: This "SLI_SI917" need to be replaced with "SLI_SI91X_MCU_INTERFACE" in other places also?

src/platform/silabs/wifi/WifiInterfaceAbstraction.h
bhmanda-silabs: If is not initialized some times it is having garbage value and condition is passing even though DUT is not provisioned.

examples/platform/silabs/wifi/icd/ApplicationSleepManager.cpp
bhmanda-silabs: "SLI_SI917" : this macro is enabled for 917 NCP and SoC boards.

bhmanda-silabs:
> @bhmanda-silabs I'm not sure i understand the way the issue is being fixed here. We can fix the issue by always configuring BLE in the associated sleep mode which simplifies our logic.
> 
Here, DUT can go deep sleep(BLE + WiFi sleep) when it is provisioned and commissioning window is closed. DUT will be active whenever requested for high performance.
> This would also mean that all the ifdef in the ApplicationSleepManager can be removed.
Modified ifdef condition to **SLI_SI917** as these checked should not be enabled for RS9116 boards.

bhmanda-silabs:
> > @bhmanda-silabs I'm not sure i understand the way the issue is being fixed here. We can fix the issue by always configuring BLE in the associated sleep mode which simplifies our logic.
> 
> Here, DUT can go deep sleep(BLE + WiFi sleep) when it is provisioned and commissioning window is closed. DUT will be active whenever requested for high performance.
> 
> > This would also mean that all the ifdef in the ApplicationSleepManager can be removed.
> > Modified ifdef condition to **SLI_SI917** as these checked should not be enabled for RS9116 boards.

@mkardous-silabs what are the sleep states required for our matter-wifi? or any design/info on the requirements?

examples/platform/silabs/wifi/icd/ApplicationSleepManager.cpp
senthilku: Are there any other places where we need to replace this macro (common code for NCP and SoC)?

mkardous-silabs:
> > > @bhmanda-silabs I'm not sure i understand the way the issue is being fixed here. We can fix the issue by always configuring BLE in the associated sleep mode which simplifies our logic.
> > 
> > 
> > Here, DUT can go deep sleep(BLE + WiFi sleep) when it is provisioned and commissioning window is closed. DUT will be active whenever requested for high performance.
> > > This would also mean that all the ifdef in the ApplicationSleepManager can be removed.
> > > Modified ifdef condition to **SLI_SI917** as these checked should not be enabled for RS9116 boards.
> 
> @mkardous-silabs what are the sleep states required for our matter-wifi? or any design/info on the requirements?

I don't have an exhaustive list of which peripherals work in each sleep state and how these peripherals behave in each sleep state. Some of this involves prototyping and testing to see how the peripherals behave and what the power consumption looks like based on the known baselines

mkardous-silabs:
> > @bhmanda-silabs I'm not sure i understand the way the issue is being fixed here. We can fix the issue by always configuring BLE in the associated sleep mode which simplifies our logic.
> 
> Here, DUT can go deep sleep(BLE + WiFi sleep) when it is provisioned and commissioning window is closed. DUT will be active whenever requested for high performance.
> 
> > This would also mean that all the ifdef in the ApplicationSleepManager can be removed.
> > Modified ifdef condition to **SLI_SI917** as these checked should not be enabled for RS9116 boards.

My point is this class exists specifically to add a new feature - Selective Listening. If a big fix is done in this class, it does not get translated to CSA which doesn't have this logic. The CanGoToDeepSleep should not go in a class whos responsibility is to determine if the device can got to LI or DTIM based sleep.

src/platform/silabs/wifi/icd/WifiSleepManager.cpp
```diff
@@ -169,7 +169,7 @@ CHIP_ERROR WifiSleepManager::VerifyAndTransitionToLowPowerMode(PowerEvent event)
     wfx_wifi_provision_t wifiConfig;
     wfx_get_wifi_provision(&wifiConfig);
 
-    if (!(wifiConfig.ssid[0] != 0))
+    if (!(wifiConfig.ssid[0] != 0) && (mCallback && mCallback->CanGoToDeepSleep()))
```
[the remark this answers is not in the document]
bhmanda-silabs: Modified the condition.

bhmanda-silabs:
> > > > @bhmanda-silabs I'm not sure i understand the way the issue is being fixed here. We can fix the issue by always configuring BLE in the associated sleep mode which simplifies our logic.
> > > 
> > > 
> > > Here, DUT can go deep sleep(BLE + WiFi sleep) when it is provisioned and commissioning window is closed. DUT will be active whenever requested for high performance.
> > > > This would also mean that all the ifdef in the ApplicationSleepManager can be removed.
> > > > Modified ifdef condition to **SLI_SI917** as these checked should not be enabled for RS9116 boards.
> > 
> > 
> > @mkardous-silabs what are the sleep states required for our matter-wifi? or any design/info on the requirements?
> 
> I don't have an exhaustive list of which peripherals work in each sleep state and how these peripherals behave in each sleep state. Some of this involves prototyping and testing to see how the peripherals behave and what the power consumption looks like based on the known baselines



> > > @bhmanda-silabs I'm not sure i understand the way the issue is being fixed here. We can fix the issue by always configuring BLE in the associated sleep mode which simplifies our logic.
> > 
> > 
> > Here, DUT can go deep sleep(BLE + WiFi sleep) when it is provisioned and commissioning window is closed. DUT will be active whenever requested for high performance.
> > > This would also mean that all the ifdef in the ApplicationSleepManager can be removed.
> > > Modified ifdef condition to **SLI_SI917** as these checked should not be enabled for RS9116 boards.
> 
> My point is this class exists specifically to add a new feature - Selective Listening. If a big fix is done in this class, it does not get translated to CSA which doesn't have this logic. The CanGoToDeepSleep should not go in a class whos responsibility is to determine if the device can got to LI or DTIM based sleep.

Ok got it. Validated by configuring BLE and WiFi to be connected state and TA is going deep sleep state after default discovery time when it  closes the commissioning window and stop the BLE. Also post-commissioning, if AP is disconnected DUT switching to low power mode.

examples/platform/silabs/wifi/icd/ApplicationSleepManager.cpp
bhmanda-silabs: No @senthilku

mkardous-silabs (changes requested):
PR proposes several changes that do not seem necessary to fix the mentionned issue.

mkardous-silabs (approved):
Approved with the expectation that the define guards will be removed with a follow up PR with the 9116 sleep fix.

senthilku (approved):

PR #341, opened 2025-03-13
[SL-TEMP] bugfix/enum check udp endpoint

lpbeliveau-silabs:
#### Testing

Built using matter_extension (with 2025 ifc2rc2) AND built with gn (using 2024 ...)

### DIFF
--- a/src/inet/UDPEndPointImplOpenThread.cpp
+++ b/src/inet/UDPEndPointImplOpenThread.cpp
@@ -27,11 +27,6 @@
 
 #include <system/SystemPacketBuffer.h>
 
-// Temporary until we manage to update the OT version in the CI
-#ifndef OT_NETIF_THREAD_HOST
-#define OT_NETIF_THREAD_HOST OT_NETIF_THREAD
-#endif
-
 namespace chip {
 namespace Inet {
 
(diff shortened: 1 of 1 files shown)


src/inet/UDPEndPointImplOpenThread.cpp
```diff
@@ -28,7 +28,9 @@
 #include <system/SystemPacketBuffer.h>
 
 // Temporary until we manage to update the OT version in the CI
-#ifndef OT_NETIF_THREAD_HOST
+// Define a macro if OT_NETIF_THREAD_HOST is present in the enum
+#define HAS_OT_NETIF_THREAD_HOST (OT_NETIF_THREAD_HOST == OT_NETIF_THREAD_HOST)
+#ifndef HAS_OT_NETIF_THREAD_HOST
```
mkardous-silabs, line 33: the `#ifdef` ins't correct anymore. `HAS_OT_NETIF_THREAD_HOST` will always defined.
```suggestion
#if !HAS_OT_NETIF_THREAD_HOST
```

lpbeliveau-silabs: How was it able to build from CSA then? Are the new ot Libs already there?

mkardous-silabs (approved):

jmartinez-silabs (approved):

PR #342, opened 2025-03-13
[CSA-CP] Rename CLUSTER_MASK_ and ATTIBUTE_MASK defines to differentiate with …

jmartinez-silabs:
**-- Cherry Pick of https://github.com/project-chip/connectedhomeip/pull/37618 --
THIS fix multiple build warning for the cmp app due to redefinitions between Matter and Zigbee**
 
…the Zigbee ones.
Clean up some scene defines and delete the now empty header att-storage.h.

CLUSTER_MASK_... -> MATTER_CLUSTER_FLAG_...
ATTRIBUTE_MASK_... -> MATTER_ATTRIBUTE_FLAG_...
Those defines are tightly coupled with the ZAP_CLUSTER_MASK and ZAP_ATTRIBUTE_MASK macros. Those public APIs are not easily modifiable to use more adequate bitflags. It would require some rework and zap changes.

Delete unused defines MATTER_DM_SCENE_TABLE_NULL_INDEX, MATTER_DM_SCENE_TABLE_UNUSED_ENDPOINT_ID and
ZCL_SCENES_CLUSTER_MAXIMUM_NAME_LENGTH

move ZCL_SCENES_GLOBAL_SCENE_GROUP_ID and ZCL_SCENES_GLOBAL_SCENE_SCENE_ID to be const members of the ScenesServer class

Testing
No behaviour change. Only a rename and/or move of defines or constants location.
Locally built for Silabs platform.
CI to validate other builds


mkardous-silabs (approved):
How do we manage CSA-CP and SL-ONLY combos?

When this trickles down to main we will need to do the same changes to our specific apps.

jmartinez-silabs:
@mkardous-silabs This PR does not have any changes to our specific apps. so it doesn't apply here.
But for a general answers, I would say the SL-ONLY trumps the CSA-CP

If you apply a SL-ONLY from a release branch to Main, which contained a CSA-CP, the CP application will lead to no diff/drop commits on the CSA-CP part

rcasallas-silabs (approved):

PR #357, opened 2025-03-18
[SL-ONLY]Add cmp datamodel sync api

jmartinez-silabs:
This PR adds a silabs internal api to sync Zigbee datamodel to matter's on zb attribute write.
`sli_matter_af_write_attribute` is called from the zigbee stack when the generated multiprotocol mapping is available.

note: file naming conventions doesn't correspond ours/matter's but was decided with the sisdk convention

#### Testing
Commission the Zigbee Matter CMP Light to matter and zigbee network. Update on-off, level-control, color-control attributes for either network and confirm the update is also done one the counterpart datamodel.

### DIFF
--- a/examples/platform/silabs/cmp/sl-matter-attribute-storage.cpp
+++ b/examples/platform/silabs/cmp/sl-matter-attribute-storage.cpp
@@ -1,0 +1,54 @@
+/*****************************************************************************
+ * @file sl-matter-attribute-storage.cpp
+ * @brief Link zigbee datamodel attribute changes to Matter attribute storage.
+ *******************************************************************************
+ * # License
+ * <b>Copyright 2024 Silicon Laboratories Inc.
+ *www.silabs.com</b>
+ *******************************************************************************
+ *
+ * The licensor of this software is Silicon
+ *Laboratories Inc. Your use of this software is
+ *governed by the terms of Silicon Labs Master
+ *Software License Agreement (MSLA) available at
+ * www.silabs.com/about-us/legal/master-software-license-agreement.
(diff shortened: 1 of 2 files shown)


examples/platform/silabs/cmp/sl-matter-attribute-storage.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <lib/support/TypeTraits.h>
+#include <lib/support/logging/CHIPLogging.h>
+#include <protocols/interaction_model/StatusCode.h>
+#include <sl-matter-attribute-storage.h>
+#include <sl_component_catalog.h>
+
+namespace {
+// This is called from simplicity_sdk/protocol/zigbee/app/framework/util/attribute-storage.c on ZB attribute write
+// when GENERATED_MULTI_PROTOCOL_ATTRIBUTE_MAPPING and SL_CATALOG_MULTIPROTOCOL_ZIGBEE_MATTER_COMMON_PRESENT are defined.
+extern "C" sl_status_t sli_matter_af_write_attribute(uint16_t endpointId, uint32_t clusterId, uint32_t attributeId,
```
Sarthak-Shaha, line 32: Do we still need the extern guards, as the header file already has it?

jmartinez-silabs: I do because it is in an anonymous namespace and used in a c file/stack (zigbee). Removing it will cause 
`-Werror=unused-function` error

Sarthak-Shaha (approved):

chirag-silabs (approved):

PR #368, opened 2025-03-24
[SL-UP] Adding app event for the LCD update

chirag-silabs:
#### Testing

Deadlock was observed when the LCD is reading any attributes which needs a lock and unlock chipstack, in some of the places the LCD update was called from the chip task which caused a deadlock.

Adding a LCD application event, which will post the event to the application for the LCD update. 

This was found during the debugging of the issue: https://jira.silabs.com/browse/MATTER-4709

Tested locally with 917 SoC and NCP, all the LCD screen updates.

### DIFF
--- a/examples/air-quality-sensor-app/silabs/include/AppEvent.h
+++ b/examples/air-quality-sensor-app/silabs/include/AppEvent.h
@@ -19,6 +19,9 @@
 
 #pragma once
 #include <cstdint>
+#ifdef DISPLAY_ENABLED
+#include "lcd.h"
+#endif
 
 struct AppEvent;
 typedef void (*EventHandler)(AppEvent *);
@@ -41,6 +44,12 @@ struct AppEvent
         {
             uint8_t Action;
         } ButtonEvent;
+#ifdef DISPLAY_ENABLED
+        struct
+        {
+            SilabsLCD::Screen_e screen;
+        } LCDEvent;
+#endif
         struct
         {
             void * Context;
--- a/examples/air-quality-sensor-app/silabs/include/AppEvent.h
+++ b/examples/air-quality-sensor-app/silabs/include/AppEvent.h
@@ -31,6 +31,7 @@ struct AppEvent
     enum AppEventTypes
     {
         kEventType_Button = 0,
+        kEventType_LCD,
         kEventType_Timer,
         kEventType_AirQualitySensor,
         kEventType_Install,
--- a/examples/lighting-app/silabs/include/AppEvent.h
+++ b/examples/lighting-app/silabs/include/AppEvent.h
@@ -27,6 +27,7 @@ struct AppEvent
     enum AppEventTypes
     {
         kEventType_Button = 0,
+        kEventType_LCD,
         kEventType_Timer,
         kEventType_Light,
         kEventType_Install,
--- a/examples/lighting-app/silabs/include/AppEvent.h
+++ b/examples/lighting-app/silabs/include/AppEvent.h
@@ -22,6 +22,7 @@
 #ifdef DISPLAY_ENABLED
 #include "lcd.h"
 #endif
+
 struct AppEvent;
 typedef void (*EventHandler)(AppEvent *);
 
--- a/examples/lighting-app/silabs/include/AppEvent.h
+++ b/examples/lighting-app/silabs/include/AppEvent.h
@@ -41,6 +41,12 @@ struct AppEvent
         {
             uint8_t Action;
         } ButtonEvent;
+#ifdef DISPLAY_ENABLED
+        struct
+        {
+            void * screen;
+        } LCDEvent;
+#endif
         struct
         {
             void * Context;
--- a/examples/lighting-app/silabs/include/AppEvent.h
+++ b/examples/lighting-app/silabs/include/AppEvent.h
@@ -19,6 +19,9 @@
 
 #pragma once
 
+#ifdef DISPLAY_ENABLED
+#include "lcd.h"
+#endif
 struct AppEvent;
 typedef void (*EventHandler)(AppEvent *);
 
@@ -44,7 +47,7 @@ struct AppEvent
 #ifdef DISPLAY_ENABLED
         struct
         {
-            void * screen;
+            SilabsLCD::Screen_e screen;
         } LCDEvent;
 #endif
         struct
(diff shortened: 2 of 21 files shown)


examples/platform/silabs/BaseApplication.cpp
```diff
@@ -251,7 +251,10 @@ void BaseApplicationDelegate::OnCommissioningWindowClosed()
         SilabsLCD::Screen_e screen;
         slLCD.GetScreen(screen);
         VerifyOrReturn(screen == SilabsLCD::Screen_e::QRCodeScreen);
-        slLCD.SetScreen(SilabsLCD::Screen_e::DemoScreen);
```
jmartinez-silabs, line 254: Should this line be deleted? How is this done now I don't see it

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -935,8 +941,10 @@ void BaseApplication::OnPlatformEvent(const ChipDeviceEvent * event, intptr_t)
         // Update the LCD screen with SSID and connected state
         if (screen == SilabsLCD::Screen_e::StatusScreen)
         {
-            BaseApplication::UpdateLCDStatusScreen(false);
-            AppTask::GetLCD().SetScreen(screen);
```
jmartinez-silabs, line 939: Again where is the SetScreen done now?

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -935,8 +941,10 @@ void BaseApplication::OnPlatformEvent(const ChipDeviceEvent * event, intptr_t)
...
+            AppEvent event;
+            event.Type    = AppEvent::kEventType_LCD;
+            event.Handler = AppTask::GetAppTask().UpdateDisplayHandler;
+            PostEvent(&event);
```
jmartinez-silabs, line 947: Should we do a `PostUpdateDisplayEvent() `function if we repeat it a few times

examples/platform/silabs/BaseApplication.cpp
chirag-silabs: Adding a variable to pass with the AppEvent, the SetScreen is done a part of the UpdateDisplay-> cycle screen

chirag-silabs: check, whether we got any data if it is nullptr cycle screen

examples/platform/silabs/BaseApplication.cpp
chirag-silabs: Passing the screen as a data in the app event

examples/platform/silabs/BaseApplication.cpp
chirag-silabs: The point if an function is introduced the event data can not be futhur changed so it would be better in my opinion to post it as it is, since this way we can pass the required event data if needed.

examples/lighting-app/silabs/include/AppEvent.h
```diff
@@ -40,6 +41,12 @@ struct AppEvent
         {
             uint8_t Action;
         } ButtonEvent;
+#ifdef DISPLAY_ENABLED
+        struct
+        {
+            void * screen;
```
jmartinez-silabs, line 47: Why is this a void pointer?

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -251,7 +251,11 @@ void BaseApplicationDelegate::OnCommissioningWindowClosed()
         SilabsLCD::Screen_e screen;
         slLCD.GetScreen(screen);
         VerifyOrReturn(screen == SilabsLCD::Screen_e::QRCodeScreen);
-        slLCD.SetScreen(SilabsLCD::Screen_e::DemoScreen);
+        AppEvent event;
+        event.Type            = AppEvent::kEventType_LCD;
+        event.LCDEvent.screen = reinterpret_cast<void *>(static_cast<uintptr_t>(SilabsLCD::Screen_e::DemoScreen));
```
jmartinez-silabs, line 256: Why don't we use the Screen_e enum type here rather than all those casting?

examples/platform/silabs/display/lcd.h
```diff
@@ -40,6 +40,7 @@ class SilabsLCD
         QRCodeScreen,
 #endif
         InvalidScreen,
+        CycleScreen,
```
jmartinez-silabs, line 43: InvalidScreen should be the last entry of the enum

examples/platform/silabs/BaseApplication.cpp
jmartinez-silabs: I am not sure if I understood exactly what you meant, but from my code review, we would just have to pass as a argument to this postLcdUpdateEvent the Screen_e enum value

chirag-silabs: Added the PostUpdateDisplayEvent

examples/lighting-app/silabs/include/AppEvent.h
chirag-silabs: moved the type to Screen_e

examples/platform/silabs/BaseApplication.cpp
chirag-silabs: done

examples/platform/silabs/display/lcd.h
chirag-silabs: Done

jmartinez-silabs:
This requires an SL-UP tag/prefix

EtienneCampeauSilabs:
Please add the required description. This PR description is not explaining the reason for the change and the Testing description is missing details.

chirag-silabs:
> Please add the required description. This PR description is not explaining the reason for the change and the Testing description is missing details.

Sorry, forgot to add. It was created a draft PR so missed to update the description back, before posting it for a review

examples/air-quality-sensor-app/silabs/include/AppEvent.h
```diff
@@ -19,6 +19,9 @@
 
 #pragma once
 #include <cstdint>
+#ifdef DISPLAY_ENABLED
+#include "lcd.h"
+#endif
 
 struct AppEvent;
```
jmartinez-silabs, line 26: This change in AppEvent.h for all apps makes me realize that there are duplications and that we should have a base AppEvent Struct. Each app should inherit the base AppEvent and add the particularity after the inheritance.

I'm not sure how/where this would be defined but we could reduce this duplication in all apps headers

chirag-silabs: yes, I have already planned to add this in CSA, since it is just a patch release don't want to add any new files. I can add the same changes here as well if needed.

jmartinez-silabs (approved):

arun-silabs (approved):

PR #377, opened 2025-03-27
[SL-UP] Adding the Custom UI for the window app

chirag-silabs:
#### Description
The window app QR code doesn't exists hence once commissioning is done the LCD display shows, no image found.
Adding the window app LCD display as a custom UI so the LCD is updated based on the UI.

JIRA: https://jira.silabs.com/browse/MATTER-4792 (image attached)

#### Testing
Tested getting the window screen after commissioning

**Note:** Noticed that there is no cycle screen present, hence the permanent fix would be adding and verifying all the btn changes and update display. On btn click as well LCD update is not happening as expected.
 Will be creating a ticket to fix this.

### DIFF
--- a/examples/window-app/silabs/src/WindowManager.cpp
+++ b/examples/window-app/silabs/src/WindowManager.cpp
@@ -171,7 +171,9 @@ void WindowManager::DispatchEventAttributeChange(chip::EndpointId endpoint, chip
     case Attributes::CurrentPositionLiftPercent100ths::Id:
     case Attributes::CurrentPositionTiltPercent100ths::Id:
         UpdateLED();
+#ifdef DISPLAY_ENABLED
         UpdateLCD();
+#endif // DISPLAY_ENABLED
         break;
     default:
         break;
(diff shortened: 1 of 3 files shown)


chirag-silabs:
Same here, this PR was also closed. Hence reopening

examples/window-app/silabs/src/WindowManager.cpp
```diff
@@ -632,10 +634,15 @@ void WindowManager::UpdateLED()
     }
 }
 
+#ifdef DISPLAY_ENABLED
+void WindowManager::DrawUI(GLIB_Context_t * glibContext)
```
senthilku, line 638: What is the use of **glibContext** param?

chirag-silabs: the callback needs it

Sarthak-Shaha (approved):

jmartinez-silabs (approved):

PR #380, opened 2025-03-28
[SL-TEMP] Increasing the CHIP Task priority to 30

chirag-silabs:
#### Description
UART Logging have more priority than the chip task causing the delay in CHIP operations when more logs are pushed to UART.
Increasing the chip task priority to 30 i.e., just more than the UART logging so that the chip task events are taken in priority than the UART

image

JIRA: https://jira.silabs.com/browse/MATTER-4730

#### Testing
Tested locally with the 917 SoC, hue value is updated in the given amount of time.


chirag-silabs:
Not sure why this PR was closed?
@mkardous-silabs was it closed due to the deletion of the release branch?

src/platform/silabs/CHIPDevicePlatformConfig.h
```diff
@@ -132,6 +132,11 @@
 #endif
 #endif // CHIP_DEVICE_CONFIG_CHIP_TASK_STACK_SIZE
 
+// Increasing the CHIP task priority more than UART(logging)
```
selissia, line 135: I'm not sure we need this comment. If this change will eventually be picked up from the CSA repo (which does not have a comment on this line) then it's better to omit it, this will simplify the merge.

selissia (approved):

src/platform/silabs/CHIPDevicePlatformConfig.h
bhmanda-silabs: On CSA, chip-task priority is 40 and here in release branch moving to 30(higher than)  according to the app task priority and UART priority. This is a temporary change is for release.

bhmanda-silabs (approved):

jmartinez-silabs (approved):

PR #385, opened 2025-03-31
[SL-UP] Brd4357a config changes fix

arun-silabs:
#### Description

For 917NCP boards, we need to generate separate binaries for BRD4346a and BRD4357a boards due to config changes between both the boards. Added the required config changes for BRD4357a under the respective macros

#### Testing
Tested the build and functionality of 917NCP boards locally.


src/platform/silabs/wifi/SiWx/WifiInterface.cpp
```diff
@@ -112,11 +112,11 @@ sl_wifi_security_t security = SL_WIFI_SECURITY_UNKNOWN;
 
 // TODO : Temporary work-around for wifi-init failure in 917NCP ACX module board(BRD4357A). Can be removed after
 // Wiseconnect fixes region code for all ACX module boards.
-#ifdef EXP_BOARD
+#ifdef ACX_MODULE_BOARD
 #define REGION_CODE IGNORE_REGION
 #else
 #define REGION_CODE US
-#endif // EXP_BOARD
+#endif // ACX_MODULE_BOARD
```
mkardous-silabs, line 119: I was under the impression that this would be fixed by the Wi-Fi SDK. Why do we still need this?

src/platform/silabs/rs911x/rsi_ble_config.h
```diff
@@ -281,15 +281,20 @@ extern "C" {
 #define FRONT_END_SWITCH_CTRL SL_SI91X_EXT_FEAT_FRONT_END_SWITCH_PINS_ULP_GPIO_4_5_0
 #endif
 
+#if (USE_BYPASS_CLOCK == 1)
+#define SL_SI91X_CLK SL_SI91X_EXT_FEAT_XTAL_CLK_ENABLE(2)
+#else
+#define SL_SI91X_CLK SL_SI91X_EXT_FEAT_XTAL_CLK_ENABLE(1)
+#endif
```
mkardous-silabs, line 288: There are already seems to be defines to do this in the Wi-Fi SDK, `SL_SI91X_EXT_FEAT_XTAL_CLK`, why do we need to have these defines in the Matter SDK?

src/platform/silabs/wifi/SiWx/WifiInterface.cpp
arun-silabs: This is fixed for 917SOC, yet to be fixed for 917NCP.

src/platform/silabs/rs911x/rsi_ble_config.h
arun-silabs: Added `SL_SI91X_CLK` to add the clock configuration in custom feature bitmap (Line 293). We should add SL_SI91X_EXT_FEAT_XTAL_CLK(1) or  SL_SI91X_EXT_FEAT_XTAL_CLK(2) as per our requirement.

src/platform/silabs/wifi/SiWx/WifiInterface.cpp
mkardous-silabs: Did they provide a reason why it was not fixed?

src/platform/silabs/rs911x/rsi_ble_config.h
mkardous-silabs: The goal of the comment is to avoid adding NCP specific configurations that would force to specify which NCP is being built.

The proposed change require changes to how the NCP are built which we should try to avoid as much as possible.

src/platform/silabs/wifi/SiWx/WifiInterface.cpp
arun-silabs: While building an application for 917NCP, the board name of the NCP carrier board will not be used, the project will be built using the efr32 board name (brd4186c or brd4187c). Due to this, an additional define for BRD4357a is not being added. 
In 917SOC, a component with board name is getting added which allows to differentiate between radio and module boards, where as this is still not done for 917NCP.

mkardous-silabs: Thank you for the background! Do we plan on supporting the NCP module in GN builds?

src/platform/silabs/rs911x/rsi_ble_config.h
mkardous-silabs: https://github.com/SiliconLabsSoftware/matter_sdk/pull/385#discussion_r2021589697 Answers the question as too why this is necessary.

src/platform/silabs/wifi/SiWx/WifiInterface.cpp
arun-silabs: Yes, will raise a PR in CSA to add a new gn build argument while building for BRD4357a NCP board

chirag-silabs (approved):

src/platform/silabs/wifi/SiWx/WifiInterface.cpp
arun-silabs: pushed the required changes for building with GN system as well.

jmartinez-silabs (approved):

mkardous-silabs:
Please don't merge the PR - would like to review

scripts/examples/gn_silabs_example.sh
```diff
@@ -203,6 +205,10 @@ else
                 shift
                 shift
                 ;;
+            --brd4357a)
+                optArgs+="use_brd4357a=true "
+                shift
+                ;;
```
mkardous-silabs, line 211: A different approach could be to change the `--wifi SiWx917` with `--wifi brd4357a` (and the non module board as well).

arun-silabs: Using `--wifi SiWx917` makes it clear that a 917 WiFi project is being built. While we can add `brd4346a` and `brd4357a` as parameters, they don't explicitly indicate that it's a 917 WiFi project unless the user is familiar with the 917NCP board names.

Additionally, if a new 917NCP radio board is introduced that doesn't require config changes, the script would need to be updated. Using `--wifi SiWx917` avoids this issue and improves clarity.

Please let me know if we should go ahead and change the --wifi argument

jmartinez-silabs: @arun-silabs 
> they don't explicitly indicate that it's a 917 WiFi project unless the user is familiar with the 917NCP board names.

That is not any different from the main board argument, which has always been BRDXXXX, and the user is supposed to know what that BRDXXXX is.

I personally like the concept proposed by @mkardous-silabs. I think it could be expanded upon. 
We could create --ncp or --rcp or something alike and pass the extension board in an argument. That would be useful for any wireless co-processor we would want to add.

In silabs_board, we would declare an argument `ncp_board` and add the logic through the gn files with that.

chirag-silabs: @arun-silabs maybe just have --wifi "module/non_module" you can come up with the name will be better

like --wifi SiWx917 and --wifi SiWx917_module for the module board. Will not prefer using the board name directly since any other boards comes in future release an addition parameter option will be needed where it can use the module/non-module board directly

arun-silabs: okay..
@mkardous-silabs @jmartinez-silabs , should we go ahead with `--wifi SiWx917` and `--wifi SiWx917_module`

mkardous-silabs: I'm not sure I agree with this approach. This keeps the hardcoded nature of only allowing wifi rpc modules which might not always be the case.

Like Junior said, the concept of BRDXXXX is already what we use for building boards in general so using the same concept for NCPs make sense.

I think the better approach is to change the --wifi logic to --rcp BRDXXXX which configure gn arg, use_x_ncp (or some version of that) which in the .gn/.gni configures the correct elements.

That being said, I am ok with this being an SL-TEMP due to the release timeline and refactor the build argument when upstreaming the changes to CSA.

PR #403, opened 2025-04-09
[SL-ONLY] Fix Apple Keychain fabric edge-cases for multi-admin

mkardous-silabs:
#### Description
While doing multi-admin test cases, I noticed while using the Google Home app on my iphone that the Appley Keychain fabric was being added to the device even if the operating fabric was not Apple. This behavior was preventing the device from entering Listen based sleep because the exception handler tried to find an Apple fabric to validate that it had an activate subscription while the fabric was not present on the device.

The PR modifies the exception handler to verify if we find an Apple fabric. If the Apple fabric is absent but we have an Apple Keychain fabric, this means that the operationnal fabric is not Apple's and the exception handler should not prevent the device from going to sleep.

#### Tests
- Updated the unit-tests to reflect the new behavior
- The multi-admin test case that allowed me to see the issue worked (Commission samsung and share to google on an Iphone)

### DIFF
--- a/examples/platform/silabs/wifi/icd/vendor-handlers/AppleKeychainHandler.h
+++ b/examples/platform/silabs/wifi/icd/vendor-handlers/AppleKeychainHandler.h
@@ -35,6 +35,8 @@ namespace Silabs {
  * This method checks if there is any fabric with the Apple Home vendor ID that
  * has at least one active subscription. If such a fabric is found, it allows
  * the device to go to LI based sleep.
+ * If the device has an Apple Keychain fabric but no Apple fabric, the device can go to LI based
+ * sleep.
  */
 class AppleKeychainHandler : public VendorHandler<AppleKeychainHandler>
 {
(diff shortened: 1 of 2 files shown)


examples/platform/silabs/wifi/icd/vendor-handlers/AppleKeychainHandler.h
```diff
@@ -44,15 +46,26 @@ class AppleKeychainHandler : public VendorHandler<AppleKeychainHandler>
     {
         VerifyOrReturnValue(subscriptionsInfoProvider != nullptr && fabricTable != nullptr, false);
 
-        bool hasValidException = false;
+        bool hasValidException = true; // Default to true if no VendorId::Apple fabric is found
 
-        for (auto it = fabricTable->begin(); it != fabricTable->end(); ++it)
+        if (fabricTable->FabricCount() == 0)
```
jmartinez-silabs, line 51: Could be `VerifyOrReturnValue(fabricTable->FabricCount(), false)`

jmartinez-silabs (approved):

examples/platform/silabs/wifi/icd/vendor-handlers/AppleKeychainHandler.h
mkardous-silabs: updated

PR #411, opened 2025-04-15
[SL-UP] Feature/common token manager

lpbeliveau-silabs:
Cleanup of SilabsConfig and implementation of common token manager API, had to keep some nvm3 calls due to lack of API for those.

#### Testing

Compile -> Provision -> Flash -> Commission -> Toggle

Compared tokens fetched with old and new API, compatibility seems to work.

### DIFF
--- a/examples/lighting-app/silabs/src/AppTask.cpp
+++ b/examples/lighting-app/silabs/src/AppTask.cpp
@@ -41,8 +41,6 @@
 #include <platform/CHIPDeviceLayer.h>
 #include <platform/silabs/tracing/SilabsTracingMacros.h>
 
-#include <platform/silabs/SilabsConfig.h>
-
 #ifdef SL_CATALOG_SIMPLE_LED_LED1_PRESENT
 #define LIGHT_LED 1
 #else
--- a/examples/lighting-app/silabs/src/AppTask.cpp
+++ b/examples/lighting-app/silabs/src/AppTask.cpp
@@ -41,6 +41,8 @@
 #include <platform/CHIPDeviceLayer.h>
 #include <platform/silabs/tracing/SilabsTracingMacros.h>
 
+#include <platform/silabs/SilabsConfig.h>
+
 #ifdef SL_CATALOG_SIMPLE_LED_LED1_PRESENT
 #define LIGHT_LED 1
 #else
--- a/examples/platform/silabs/SoftwareFaultReports.cpp
+++ b/examples/platform/silabs/SoftwareFaultReports.cpp
@@ -83,6 +83,8 @@ void OnSoftwareFaultEventHandler(const char * faultRecordString)
 } // namespace DeviceLayer
 } // namespace chip
 
+// This method is already implemented in the Zigbee stack and is required by the Zigbee
+#ifndef SL_CATALOG_ZIGBEE_STACK_COMMON_PRESENT
 extern "C" void halInternalAssertFailed(const char * filename, int linenumber)
 {
     char faultMessage[kMaxFaultStringLen] = { 0 };
@@ -92,6 +94,7 @@ extern "C" void halInternalAssertFailed(const char * filename, int linenumber)
 #endif
     configASSERT((volatile void *) NULL);
 }
+#endif
 
 #if HARD_FAULT_LOG_ENABLE
 /**
--- a/examples/platform/silabs/SoftwareFaultReports.cpp
+++ b/examples/platform/silabs/SoftwareFaultReports.cpp
@@ -83,6 +83,16 @@ void OnSoftwareFaultEventHandler(const char * faultRecordString)
 } // namespace DeviceLayer
 } // namespace chip
 
+extern "C" void halInternalAssertFailed(const char * filename, int linenumber)
+{
+    char faultMessage[kMaxFaultStringLen] = { 0 };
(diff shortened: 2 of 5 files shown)


examples/platform/silabs/SoftwareFaultReports.cpp
```diff
@@ -83,6 +83,16 @@ void OnSoftwareFaultEventHandler(const char * faultRecordString)
 } // namespace DeviceLayer
 } // namespace chip
 
+extern "C" void halInternalAssertFailed(const char * filename, int linenumber)
```
jmartinez-silabs, line 86: Does that work with cmp also ? Think it exist in zigbee stack

src/platform/silabs/SilabsConfigCTM.cpp
```diff
@@ -0,0 +1,424 @@
...
+namespace DeviceLayer {
+namespace Internal {
+namespace {
+CHIP_ERROR MapNvm3Error(uint32_t nvm3Res)
+{
+    CHIP_ERROR err;
+
+    switch (nvm3Res)
+    {
+    case ECODE_NVM3_OK:
```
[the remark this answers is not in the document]
lpbeliveau-silabs: They get cast a sl_status, which is essentially another uint32_t type.
The bottom line is, under the hood they are the same calls to nvm3 methods so we can handle the errors the same way.

examples/lighting-app/silabs/src/AppTask.cpp
```diff
@@ -41,6 +41,8 @@
 #include <platform/CHIPDeviceLayer.h>
 #include <platform/silabs/tracing/SilabsTracingMacros.h>
 
+#include <platform/silabs/SilabsConfig.h>
```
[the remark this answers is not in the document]
lpbeliveau-silabs: That one slipped through, I added this just to run my test function after the AppInit was completed, this should have been removed.

src/platform/silabs/SilabsConfig.cpp
```diff
@@ -62,6 +62,59 @@ namespace chip {
 namespace DeviceLayer {
 namespace Internal {
 
+namespace {
```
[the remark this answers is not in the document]
lpbeliveau-silabs: Not unless we want the whole definition in the .h file OR to add:
```
template CHIP_ERROR SilabsConfig::ReadConfigValue<bool>(Key key, bool & val);
template CHIP_ERROR SilabsConfig::ReadConfigValue<uint16_t>(Key key, uint16_t & val);
template CHIP_ERROR SilabsConfig::ReadConfigValue<uint32_t>(Key key, uint32_t & val);
template CHIP_ERROR SilabsConfig::ReadConfigValue<uint64_t>(Key key, uint64_t & val);
```
In the cpp file.

I do not have a preference though to be honest.

jmartinez-silabs (approved):

examples/platform/silabs/SoftwareFaultReports.cpp
```diff
@@ -83,6 +83,19 @@ void OnSoftwareFaultEventHandler(const char * faultRecordString)
 } // namespace DeviceLayer
 } // namespace chip
 
+// This method is already implemented in the Zigbee stack and is required by the Zigbee
+#ifndef SL_CATALOG_ZIGBEE_STACK_COMMON_PRESENT
+extern "C" void halInternalAssertFailed(const char * filename, int linenumber)
+{
+    char faultMessage[kMaxFaultStringLen] = { 0 };
+    snprintf(faultMessage, sizeof faultMessage, "Assert failed: %s:%d", filename, linenumber);
+#if SILABS_LOG_ENABLED
```
jepenven-silabs, line 92: Shouldn't this be move on top of the faultMessage declaration ? I mean if the logs are not enabled, the buffer declaration and assignment seems pointless...

jepenven-silabs (approved):

mkardous-silabs:
Force merging since the PR doesn't affect ZAP and ZAP workflow is not yet fixed.

src/platform/silabs/SilabsConfig.h
```diff
@@ -201,15 +223,13 @@ class SilabsConfig
...
-    static CHIP_ERROR ReadConfigValueBin(Key key, uint8_t * buf, size_t bufSize, size_t & outLen, size_t offset);
+    static CHIP_ERROR ReadConfigValueBin(Key key, uint8_t * buf, size_t bufSize, size_t & outLen, size_t offset = 0);
     static CHIP_ERROR ReadConfigValueCounter(uint8_t counterIdx, uint32_t & val);
     static CHIP_ERROR WriteConfigValue(Key key, bool val);
     static CHIP_ERROR WriteConfigValue(Key key, uint16_t val);
     static CHIP_ERROR WriteConfigValue(Key key, uint32_t val);
     static CHIP_ERROR WriteConfigValue(Key key, uint64_t val);
-    static CHIP_ERROR WriteConfigValueStr(Key key, const char * str);
-    static CHIP_ERROR WriteConfigValueStr(Key key, const char * str, size_t strLen);
+    static CHIP_ERROR WriteConfigValueStr(Key key, const char * str, size_t strLen = 0);
```
rosahay-silabs, line 232: `strLen` should not be set to 0 by default right?

lpbeliveau-silabs: This allows for combining the API's into one signature. We do need to import the sizing logic though.

PR #424, opened 2025-04-30
[SL-ONLY] Update csa update workflows to use github app

mkardous-silabs:
### Description
In the initial implementations of the automation workflows used a PAT (personal access token) associated to my github account for the missing permissions the `GITHUB_TOKEN` available by default did not have. The issue with the PAT is that it executed actions with my github handle at moments where it would be preferable to have a "bot" do them.

By replacing the PAT with a github app, the permissions are associated with the github app and not to my personnal account. This removes the depency on my to keep everything going and simplifies some of the automation since we don't need to work around the lacking permissions of the github token.

PR replaces the PAT where it is used in the matter_sdk by the github app token. The changes to use the Github App is only done where the github token does not have sufficient permissions to avoid having operation executed wihtout realising it. Leveraging the Github App should be deliberate for now.

### Tests
- Open PR test: https://github.com/SiliconLabsSoftware/matter_sdk/pull/423
- CSA update test: https://github.com/SiliconLabsSoftware/matter_sdk/actions/runs/14762349836/workflow

### DIFF
--- a/.github/workflows/silabs-open-csa-pr.yaml
+++ b/.github/workflows/silabs-open-csa-pr.yaml
@@ -16,9 +16,18 @@ jobs:
         if: ${{ github.event.workflow_run.conclusion == 'success' }}
         runs-on: ubuntu-latest
         steps:
+            - name: Generate token for the workflow
+              id: generate_token
+              uses: actions/create-github-app-token@v2
+              with:
+                  app-id: ${{ vars.SILABSSW_MATTER_CI_BOT_APP_ID }}
+                  private-key:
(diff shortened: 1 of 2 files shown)


mkardous-silabs:
The failures for the closure-app and the refrigerator-app are expected. The closure-app does not yet build and is waiting on PRs to be merged on CSA.

Refrigerator fix was merged in CSA and will tickled down in the next update from csa.

jmartinez-silabs (approved):

Sarthak-Shaha (approved):

mkardous-silabs:
Merging PR since the remaining jobs are unrelated and we require an admin merge due to the Refrigerator failure.

PR #427, opened 2025-04-30
[SL-ONLY] Replace javascript exit call with bash exit

mkardous-silabs:
### Description 
Exit call being used was javascript code in a bash context. PR replaces with the correct exit function.

### Tests
See CI failure - doesn't crash anymore.

Failure case : https://github.com/SiliconLabsSoftware/matter_sdk/actions/runs/14764010735/job/41451227702


jmartinez-silabs (approved):

selissia (approved):

mkardous-silabs:
Merging since the changes don't impact the on-going CI and is necessary to the refregirator failure.

PR #434, opened 2025-05-03
[SL-UP] Set correct length for Wi-Fi SSID and PSK in WriteConfigValueStr

senthilku:
#### Problem
https://jira.silabs.com/browse/MATTER-4930
Wi-Fi devices fail to connect to the access point (AP) after a power cycle or OTA update due to the following reason,
The SilabsConfig cleanup PR (https://github.com/SiliconLabsSoftware/matter_sdk/pull/411) removed the WriteConfigValueStr(Key key, const char * str) API, which was used to store the SSID and passkey for Wi-Fi. This change is affecting all the Wi-Fi devices.

#### Solution
Updated the code to use SilabsConfig::WriteConfigValueStr(Key key, const char * str, size_t strLen) for storing Wi-Fi credentials

#### Testing
Tested commissioning and verified reconnection after a power cycle using the BRD4338A board.


rosahay-silabs:
I am a bit confused, if this API was deprecated then it should not have found the said API in the #411, are we missing something here?

src/platform/silabs/NetworkCommissioningWiFiDriver.cpp
```diff
@@ -75,8 +75,10 @@ CHIP_ERROR SlWiFiDriver::CommitConfiguration()
 {
     uint8_t securityType = WFX_SEC_WPA2;
 
-    ReturnErrorOnFailure(SilabsConfig::WriteConfigValueStr(SilabsConfig::kConfigKey_WiFiSSID, mStagingNetwork.ssid));
-    ReturnErrorOnFailure(SilabsConfig::WriteConfigValueStr(SilabsConfig::kConfigKey_WiFiPSK, mStagingNetwork.credentials));
+    ReturnErrorOnFailure(
+        SilabsConfig::WriteConfigValueStr(SilabsConfig::kConfigKey_WiFiSSID, mStagingNetwork.ssid, strlen(mStagingNetwork.ssid)));
```
rosahay-silabs, line 79:
```suggestion
        SilabsConfig::WriteConfigValueStr(SilabsConfig::kConfigKey_WiFiSSID, mStagingNetwork.ssid, mStagingNetwork.ssidLen));
```

rosahay-silabs (changes requested):

senthilku:
> I am a bit confused, if this API was deprecated then it should not have found the said API in the #411 CICD runs and have been caught there, are we missing something here?

These two methods are overloaded functions for string writing, and their signatures are the same.
It won't cause any issues in CI/CD or local builds.
static CHIP_ERROR WriteConfigValueStr(Key key, const char * str); - It is deprcated
static CHIP_ERROR WriteConfigValueStr(Key key, const char * str, size_t strLen);

rosahay-silabs (approved):

mkardous-silabs (approved):

mkardous-silabs:
Open / Close the PR to kick start the CI

PR #435, opened 2025-05-05
[SL-ONLY] Adding the LCD event for silabs only app

chirag-silabs:
#### Testing

Refactoring the AppEvent to a common place.
Adding an LCD App event when the LCD update is needed. This makes sure that there will be no deadlock from the chip task when LCD update is done from the Chip Task which needs it internally to lock chip task i.e., get any attribute data.


examples/multi-sensor-app/silabs/src/DataModelCallbacks.cpp
```diff
@@ -64,10 +64,9 @@ void MatterPostAttributeChangeCallback(const chip::app::ConcreteAttributePath &
         VerifyOrReturn(attributeId == TemperatureMeasurement::Attributes::MeasuredValue::Id ||
                        attributeId == RelativeHumidityMeasurement::Attributes::MeasuredValue::Id);
 
-        AppEvent event = {
-            .Type    = AppEvent::kEventType_SensorAttributeUpdate,
-            .Handler = AppTask::SensorAttributeUpdateEvent,
-        };
+        AppEvent event = {};
+        event.Type     = AppEvent::kEventType_SensorAttributeUpdate;
+        event.Handler  = AppTask::SensorAttributeUpdateEvent;
```
mkardous-silabs, line 69: This change isn't necessary.

chirag-silabs: it was needed, that's why the multi-sensor build was failng earlier on the CSA sync branch, when you cherrypicked the changes.

mkardous-silabs: Why is it needed? In theory, you should be able to configure the AppEvent directly in the instatiation.

PR #436, opened 2025-05-05
[SL-UP] Sizeless WriteConfigValueStr fix

lpbeliveau-silabs:
Added the logic to get the size from the string when not provided in WriteConfigValueStr

#### Testing

Build, flash, commission, toggle

### DIFF
--- a/src/platform/silabs/SilabsConfig.cpp
+++ b/src/platform/silabs/SilabsConfig.cpp
@@ -284,6 +284,11 @@ CHIP_ERROR SilabsConfig::WriteConfigValueStr(Key key, const char * str, size_t s
 
     if (str != NULL)
     {
+        // Compute the length of the string if not provided.
+        if (strLen == 0)
+        {
+            strLen = strlen(str);
+        }
         // Write the string to nvm3 without the terminator char (apart from
         // empty strings where only the terminator char is stored in nvm3).
         err = MapNvm3Error(nvm3_writeData(nvm3_defaultHandle, key, str, (strLen > 0) ? strLen : 1));
--- a/src/platform/silabs/SilabsConfig.h
+++ b/src/platform/silabs/SilabsConfig.h
@@ -229,6 +229,16 @@ public:
     static CHIP_ERROR WriteConfigValue(Key key, uint16_t val);
     static CHIP_ERROR WriteConfigValue(Key key, uint32_t val);
     static CHIP_ERROR WriteConfigValue(Key key, uint64_t val);
+    /** @brief WriteConfigValueStr
+     *  Write a string to the NVM3 storage. This function can be called without providing the length of the string.
+     *  In that case, the length is calculated using strlen() function. Note, this is unsafe as it can lead to buffer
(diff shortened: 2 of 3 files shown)


src/platform/silabs/SilabsConfig.cpp
```diff
@@ -284,6 +284,11 @@ CHIP_ERROR SilabsConfig::WriteConfigValueStr(Key key, const char * str, size_t s
 
     if (str != NULL)
```
mkardous-silabs, line 285: Can we use the verify or return logic here as well?

mkardous-silabs (approved):

src/platform/silabs/SilabsConfig.cpp
```diff
@@ -284,6 +284,11 @@ CHIP_ERROR SilabsConfig::WriteConfigValueStr(Key key, const char * str, size_t s
...
     {
+        // Compute the length of the string if not provided.
+        if (strLen == 0)
+        {
+            strLen = strlen(str);
```
jmartinez-silabs, line 290: If str is not null-terminated, then this will cause out of bound read and write garbage to nvm3 possibly a large amount of data.

also, not to important but if the user did provide a strLen of 0 and it isn't the default value, then we are doing strlen for no reason but that not to dramatic.

src/platform/silabs/SilabsConfig.cpp
lpbeliveau-silabs: Yes but I didn't want to go through the whole exit cleanup in our original SilabsConfig since the goal is to switch to SilabsConfigCTM (And remove CTM) once the Common Token Manager is in a useable state.

The cleanup of SilabsConfig is a task in itself.

src/platform/silabs/SilabsConfig.cpp
rosahay-silabs, line 290: we should use `strnlen` instead to prevent buffer overflows right?

lpbeliveau-silabs: Yes, but we aren't providing the size in this usecase... we could define a max string size but in the event of a string smaller than max, we could still overflow.

lpbeliveau-silabs: You are correct, however the unit test are expecting a callable API without a size. I think we could configure a maximum size nvm3 object, but in the event someone writes a non-null terminated string in a buffer that is smaller than that max, we would still have the possibility for overflows.

rosahay-silabs: can it be possible it's a null string with just a key?

lpbeliveau-silabs: Strnlen requires the length, since we don't know it here, that wouldn't be much help.

lpbeliveau-silabs: we have a check on NULL 5 lines above this line.

jmartinez-silabs (approved):

PR #439, opened 2025-05-05
[SL-UP] BLE Side Channel

lpbeliveau-silabs:
Introducing ble side Channel with shell commands, POC currently doesn't support indication.

#### Testing

Build with and without the side channel on SLC and ninja.

Shell : Advertise Start/Stop in parallel to CHIPoBLE

Phone App: Connect, Read, Write


src/platform/silabs/efr32/BLEChannel.cpp
```diff
@@ -0,0 +1,319 @@
...
+#include <platform/internal/CHIPDeviceLayerInternal.h>
+
+namespace chip {
+namespace DeviceLayer {
+namespace Internal {
+
+namespace {
+
+// Side Channel UUIDS
+const uint8_t kSideServiceUUID[16] = { 0x01, 0x00, 0x00, 0xEE, 0xFF, 0xC0, 0x00, 0x80,
```
jmartinez-silabs, line 14: how are those uuid decide? Could a user want to set those?
Maybe we can use defines for them and use
ifndef
define

src/platform/silabs/efr32/BLEManagerImpl.cpp
```diff
@@ -109,6 +114,12 @@ const uint8_t ShortUUID_CHIPoBLEService[] = { 0xF6, 0xFF };
 
 bd_addr randomizedAddr = { 0 };
 
+bool isCHIPoBLECharacteristic(uint16_t characteristic)
```
jmartinez-silabs, line 117:
```suggestion
bool isMATTERoBLECharacteristic(uint16_t characteristic)
```

src/platform/silabs/efr32/BLEManagerImpl.cpp
```diff
@@ -663,139 +765,207 @@ void BLEManagerImpl::HandleBootEvent(void)
...
+    ChipLogProgress(DeviceLayer, "Connection Parameters Event for handle : %d", con_param_evt->connection);
+    ChipLogProgress(DeviceLayer, "Interval: %d, Latency: %d, Timeout: %d", con_param_evt->interval, con_param_evt->latency,
+                    con_param_evt->timeout);
+
+    uint16_t desiredTimeout = con_param_evt->timeout < BLE_CONFIG_TIMEOUT ? BLE_CONFIG_TIMEOUT : con_param_evt->timeout;
+
+    // For better stability, renegotiate the connection parameters if the received ones from the central are outside
+    // of our defined constraints
+    if (desiredTimeout != con_param_evt->timeout || con_param_evt->interval < BLE_CONFIG_MIN_INTERVAL ||
+        con_param_evt->interval > BLE_CONFIG_MAX_INTERVAL)
```
jmartinez-silabs, line 804: I'd argue that connection intervals have not been an issue in the past. Only the timeout. This might "slow" the ble communication (and connection establishment) unnecessarily.

src/platform/silabs/efr32/BLEManagerImpl.cpp
```diff
@@ -1123,8 +1293,7 @@ extern "C" void sl_bt_on_event(sl_bt_msg_t * evt)
             chip::DeviceLayer::Internal::BLEMgrImpl().HandleTxConfirmationEvent(
                 evt->data.evt_gatt_server_characteristic_status.connection);
         }
-        else if ((evt->data.evt_gatt_server_characteristic_status.characteristic == gattdb_CHIPoBLEChar_Tx) &&
-                 (evt->data.evt_gatt_server_characteristic_status.status_flags == sl_bt_gatt_server_client_config))
+        else
```
jmartinez-silabs, line 1296: how do we let the User of the side channel handle other possible event they might be interested on.

With this implementation we don't let them handle any

src/platform/silabs/BLEManagerImpl.h
```diff
@@ -32,6 +32,7 @@
 #include "gatt_db.h"
 #include "sl_bgapi.h"
 #include "sl_bt_api.h"
+#include <BLEChannel.h>
```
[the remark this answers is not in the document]
lpbeliveau-silabs: I thought about it, I'd rather just leave a pointer for a potential side channel in the object.

Otherwise we need to ifedf every checks on `VerifyOrReturn(mBleSideChannel != nullptr);`

src/platform/silabs/efr32/BLEChannel.cpp
lpbeliveau-silabs: A user would implement their own BLEChannelImpl.cpp object. I just realized this should instead be named:
BLEChannelSideImpl.cpp or something like that.

src/platform/silabs/efr32/BLEManagerImpl.cpp
lpbeliveau-silabs: This came from somewhere else, I remember it popping up in a rebase conflict and I decided to keep it but nothing more. I don't mind reverting it after I confirm where it came from.

src/platform/silabs/efr32/BLEManagerImpl.cpp
lpbeliveau-silabs: Do tou mean that the switch case statement should have a `default` that is handled by the side channel? If that is the case, I agree, I'll look into implementing a virtual method for handling other events in the BLEChannel object.

If this is regarding sl_bt_evt_gatt_server_characteristic_status_id, to my knowledge this is use either to recevie request for subscription to a characteristic supporting indication or to confirm indication was sent. The side channel implementation can decide what logic they want to use in the `HandleCCCDWriteRequest`

src/platform/silabs/efr32/BLEManagerImpl.cpp
jmartinez-silabs: got it

src/platform/silabs/efr32/BLEManagerImpl.cpp
```diff
@@ -663,139 +771,207 @@ void BLEManagerImpl::HandleBootEvent(void)
...
 
 void BLEManagerImpl::HandleConnectionCloseEvent(volatile sl_bt_msg_t * evt)
 {
     sl_bt_evt_connection_closed_t * conn_evt = (sl_bt_evt_connection_closed_t *) &(evt->data);
-    uint8_t connHandle                       = conn_evt->connection;
 
-    ChipLogProgress(DeviceLayer, "Disconnect Event for handle : %d", connHandle);
-
-    if (RemoveConnection(connHandle))
+    if (NULL != GetConnectionState(conn_evt->connection, false))
```
jmartinez-silabs, line 828:
```suggestion
    if (nullptr != GetConnectionState(conn_evt->connection, false))
```

src/platform/silabs/efr32/BLEManagerImpl.cpp
```diff
@@ -1153,12 +1344,31 @@ extern "C" void sl_bt_on_event(sl_bt_msg_t * evt)
...
     chip::DeviceLayer::PlatformMgr().UnlockChipStack();
 }
 
+} // namespace Internal
+} // namespace DeviceLayer
+} // namespace chip
+
+extern "C" void zigbe_bt_on_event(volatile sl_bt_msg_t * evt);
+
+extern "C" void sl_bt_on_event(sl_bt_msg_t * evt)
```
jmartinez-silabs, line 1363: We might want those somewhere else in the final implementation

src/platform/silabs/efr32/BLEManagerImpl.cpp
```diff
@@ -1153,12 +1344,31 @@ extern "C" void sl_bt_on_event(sl_bt_msg_t * evt)
...
 
+    // Unlock the stack
     chip::DeviceLayer::PlatformMgr().UnlockChipStack();
 }
 
+} // namespace Internal
+} // namespace DeviceLayer
+} // namespace chip
+
+extern "C" void zigbe_bt_on_event(volatile sl_bt_msg_t * evt);
```
jmartinez-silabs, line 1361: does this need a #ifdef SL_TEST_MATTER_ZIGBEE_DMP too

examples/platform/silabs/gatt_configuration.btconf
```diff
@@ -96,4 +96,43 @@
     </characteristic>
 
   </service>
+  
+    <!--Silabs DMP Light-->
```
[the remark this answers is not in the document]
lpbeliveau-silabs: To move it in the extension, I haven't figured out a clean slcc plan yet so I just put it in here for now but the override files location should be in extension.

src/platform/silabs/BLEManagerImpl.h
```diff
@@ -32,6 +32,8 @@
...
```
mkardous-silabs, line 35:
```suggestion
#include <platform/silabs/.../BLEChannel.h>
```

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -45,6 +45,15 @@
 #endif // CHIP_CONFIG_ENABLE_ICD_SERVER
 
 #include <app/server/OnboardingCodesUtil.h>
+
+#if SL_BLE_SIDE_CHANNEL_ENABLED
```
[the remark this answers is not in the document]
lpbeliveau-silabs: I am not sure about this, nothing about this side channel is matter specific and does not interact with anything matter related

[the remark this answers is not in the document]
lpbeliveau-silabs: Plus a Matter side channel is not a thing

src/platform/silabs/efr32/BLEChannel.h
```diff
@@ -0,0 +1,188 @@
...
+public:
+    BLEChannel()
+    {
+        // Initialize the connection state
+        memset(&mConnectionState, 0, sizeof(mConnectionState));
+        mFlags.ClearAll();
+        mAdvIntervalMin = 0;
+        mAdvIntervalMax = 0;
+        mAdvDuration    = 0;
+        mAdvMaxEvents   = 0;
```
mkardous-silabs, line 47: Is this necessary in the constructor? default values already configure this so this seems like unnecessary steps.

src/platform/silabs/efr32/BLEChannel.h
```diff
@@ -0,0 +1,188 @@
...
+     *  Determine if the side channel can handle a BLE event. This only gets called if the
+     *  targetted event is not supported by MATTERoBLE as the side channel supports every
+     *  event that MATTERoBLE does.
+     *
+     * @param event uint32_t Header of sl_bt_msg_t event type
+     * @return true if the event can be handled, false otherwise.
+     */
+    virtual bool CanHandleEvent(uint32_t event) { return false; }
+
+    virtual void ParseEvent(volatile sl_bt_msg_t * evt) {}
```
mkardous-silabs, line 86: Why are these the only functions that are not pure virtual?

third_party/silabs/efr32_sdk.gni
```diff
@@ -662,6 +665,13 @@ template("efr32_sdk") {
       }
     }
 
+    if (sl_enable_ble_side_channel) {
+      defines += [
+        "SILABS_USE_BLE_SIDE_CHANNEL=1",
+        "SL_BT_ADVERTISER_CONFIG_H",
+        "SL_BT_CONFIG_USER_ADVERTISERS=2",
```
mkardous-silabs, line 672: `SL_MATTER_*` structure

src/platform/silabs/efr32/BLEManagerImpl.cpp
```diff
@@ -118,7 +129,13 @@ CHIP_ERROR BLEManagerImpl::_Init()
     // Initialize the CHIP BleLayer.
     ReturnErrorOnFailure(BleLayer::Init(this, this, &DeviceLayer::SystemLayer()));
 
+    // This line hasn't changed but since the BLEConState definition has moved to
```
mkardous-silabs, line 132: could this be due to the fact that it finds two structure?

src/platform/silabs/efr32/BLEChannelImpl.cpp
```diff
@@ -0,0 +1,431 @@
...
+    default:
+        return CHIP_ERROR(ChipError::Range::kPlatform, bleErr + CHIP_DEVICE_CONFIG_SILABS_BLE_ERROR_MIN);
+    }
+}
+
+} // namespace
+
+CHIP_ERROR BLEChannelImpl::Init()
+{
+    uint16_t session;
```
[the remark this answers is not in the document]
lpbeliveau-silabs: I really do not see in what it makes a difference, in my opinion, this is worst because it leads the reader to think that this 0 value is used somewhere, which isn't the case.

src/platform/silabs/efr32/BLEChannelImpl.cpp
```diff
@@ -0,0 +1,431 @@
...
+
+    ret = sl_bt_gattdb_commit(session);
+    return MapBLEError(ret);
+}
+
+CHIP_ERROR BLEChannelImpl::ConfigureAdvertising(const AdvConfigStruct & config)
+{
+    ChipLogProgress(DeviceLayer, "Configuring BLE Channel");
+
+    sl_status_t ret;
```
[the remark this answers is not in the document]
lpbeliveau-silabs: Same here, there is no use for this assignment, and it lets the reader think there might be one.

src/platform/silabs/efr32/BLEManagerImpl.cpp
lpbeliveau-silabs: I'll add one just in case, but since it doesn't get calledI think the linker wouldn't complain about not finding the definition.

src/platform/silabs/efr32/BLEManagerImpl.cpp
lpbeliveau-silabs: I'll take note, like a matter_ble_event.cpp? I like the idea, I just want to avoid going through adding stuff for the gn builds for now as I would ideally not test gn builds since they take forever. I'll put a todo in there for now.

src/platform/silabs/efr32/BLEChannel.h
lpbeliveau-silabs: Because the ability to parse events in the Channel will be implemented in the future.

src/platform/silabs/BLEManagerImpl.h
```diff
@@ -69,19 +71,79 @@ class BLEManagerImpl final : public BLEManager, private BleLayer, private BlePla
...
+    CHIP_ERROR SideChannelStartAdvertising(void);
+    CHIP_ERROR SideChannelStopAdvertising(void);
+
+    // GAP
+    CHIP_ERROR SideChannelGeneratAdvertisingData(uint8_t discoverMove, uint8_t connectMode, const Optional<uint16_t> & maxEvents)
+    {
+        VerifyOrReturnError(mBleSideChannel != nullptr, CHIP_ERROR_INCORRECT_STATE);
+        return mBleSideChannel->GeneratAdvertisingData(discoverMove, connectMode, maxEvents);
+    }
+    CHIP_ERROR SideChannelOpenConnection(bd_addr address, uint8_t addrType)
```
jmartinez-silabs, line 97: NIT: spacing between functions for readability please

jmartinez-silabs (approved):

examples/platform/silabs/BaseApplication.cpp
jepenven-silabs, line 49: Pre-processor macro are not uniformed in this file, here it's #if and at line 137 you also check if the define is actually defined. Please unify the check across the file

examples/platform/silabs/shell/ble/BLEShellCommands.cpp
```diff
@@ -0,0 +1,94 @@
...
+    err = DeviceLayer::Internal::BLEMgrImpl().SideChannelStartAdvertising();
+    VerifyOrReturnError(err == CHIP_NO_ERROR, err,
+                        streamer_printf(streamer_get(), "Failed to start BLE side channel advertising: %s\n", ErrorStr(err)););
+    streamer_printf(streamer_get(), "Started BLE side channel advertising\n");
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR StopBLESideChannelAvertising(int argc, char ** argv)
+{
+    DeviceLayer::Internal::BLEMgrImpl().SideChannelStopAdvertising();
```
jepenven-silabs, line 73: I mean, StopAdvertising retourne un CHIP_ERROR, fac la fonction devrait au moins forwarder le status de ce qu'elle run right ? du genre : 

`return DeviceLayer::Internal::BLEMgrImpl().SideChannelStopAdvertising();`

examples/platform/silabs/shell/ble/BLEShellCommands.cpp
```diff
@@ -0,0 +1,94 @@
...
+    // TODO : Configure first
+    CHIP_ERROR err = DeviceLayer::Internal::BLEMgrImpl().SideChannelConfigureAdvertisingDefaultData();
+    VerifyOrReturnError(
+        err == CHIP_NO_ERROR, err,
+        streamer_printf(streamer_get(), "Failed to configure the BLE side channel advertising: %s\n", ErrorStr(err)););
+    err = DeviceLayer::Internal::BLEMgrImpl().SideChannelStartAdvertising();
+    VerifyOrReturnError(err == CHIP_NO_ERROR, err,
+                        streamer_printf(streamer_get(), "Failed to start BLE side channel advertising: %s\n", ErrorStr(err)););
+    streamer_printf(streamer_get(), "Started BLE side channel advertising\n");
+    return CHIP_NO_ERROR;
```
jepenven-silabs, line 68: Meme chose on devrait forward l'erreur a la shell qui handle deja les commands failures

src/platform/silabs/efr32/BLEChannel.h
jepenven-silabs: I would add a comment with a TODO then so that we remember this info

src/platform/silabs/efr32/BLEChannelImpl.cpp
```diff
@@ -0,0 +1,447 @@
...
+#include <platform/internal/CHIPDeviceLayerInternal.h>
+
+namespace chip {
+namespace DeviceLayer {
+namespace Internal {
+
+namespace {
+
+// Side Channel UUIDS
+const uint8_t kSideServiceUUID[16] = { 0x01, 0x00, 0x00, 0xEE, 0xFF, 0xC0, 0x00, 0x80,
```
jepenven-silabs, line 31: I'm really, but really not a big fan of all this hardcoded stuff in a CPP file. This should all be moved to some macro somewhere in the form of 

```#define SIDE_CHANNEL_UUID 0x01, 0x00, 0x00, 0xEE, 0xFF, 0xC0, 0x00, 0x80, 0x00, 0x10, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00``` 

and then 
`const uint8_t kSideServiceUUID[16] = { SIDE_CHANNEL_UUID };`

File: src/platform/silabs/efr32/BLEChannelImpl.cpp (hunk 16 above)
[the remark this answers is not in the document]
Comment (jepenven-silabs) [reply src/platform/silabs/efr32/BLEChannelImpl.cpp:64]: Opinion aside, an uninitialized variable is always a bad practice (not MISRA compliant for example) this should be fixed

File: src/platform/silabs/efr32/BLEChannelImpl.cpp [hunk 22]
```diff
@@ -0,0 +1,447 @@
...
+    ret = sl_bt_gattdb_add_uuid128_characteristic(session, mSideServiceHandle,
+                                                  SL_BT_GATTDB_CHARACTERISTIC_READ | SL_BT_GATTDB_CHARACTERISTIC_WRITE |
+                                                      SL_BT_GATTDB_CHARACTERISTIC_WRITE_NO_RESPONSE |
+                                                      SL_BT_GATTDB_CHARACTERISTIC_INDICATE,
+                                                  0, // No security
+                                                  0, // No flags
+                                                  kTxUUID, sl_bt_gattdb_variable_length_value,
+                                                  255, // Max length
+                                                  sizeof(InitialValueTX), InitialValueTX, &mSideTxCharHandle);
+    VerifyOrReturnError(ret == SL_STATUS_OK, MapBLEError(ret));
```
jepenven-silabs, line 93: There are too many calls to `MapBLEError` in the following lines, 
it would be best to change it to : 
```
VerifyOrExit(ret == SL_STATUS_OK);


exit : 
return MapBLEError(ret);
```

src/platform/silabs/efr32/BLEChannelImpl.cpp
lpbeliveau-silabs: We already have a way to associate this hardcoded value to the side_channel_uuid, it's kSideServiceUUID, adding a define would just be convoluted.

src/platform/silabs/efr32/BLEChannelImpl.cpp
```diff
@@ -0,0 +1,447 @@
...
+    return MapBLEError(ret);
+}
+
+CHIP_ERROR BLEChannelImpl::ConfigureAdvertising(const AdvConfigStruct & config)
+{
+    ChipLogProgress(DeviceLayer, "Configuring BLE Channel");
+
+    sl_status_t ret;
+
+    if (mAdvHandle == 0xff)
```
jepenven-silabs, line 114: Magic Number, should be a define

src/platform/silabs/efr32/BLEChannelImpl.cpp
lpbeliveau-silabs: We are not putting exit in our code anymore.

src/platform/silabs/efr32/BLEChannelImpl.cpp
```diff
@@ -0,0 +1,447 @@
...
+        // TODO: Check if we need to randomize the address
+        uint64_t random = Crypto::GetRandU64();
+        memcpy(&mConnectionState.address, &random, sizeof(mConnectionState.address));
+
+        // Set two MSBs to 11 to properly the address - BLE Static Device Address requirement
+        mConnectionState.address.addr[5] |= 0xC0;
+
+        ret = sl_bt_advertiser_set_random_address(mAdvHandle, sl_bt_gap_static_address, mConnectionState.address,
+                                                  &mConnectionState.address);
+        VerifyOrReturnLogError(ret == SL_STATUS_OK, MapBLEError(ret));
```
jepenven-silabs, line 128: Before it was `VerifyOrReturnError` now it's `VerifyOrReturnLogError` Why?

src/platform/silabs/efr32/BLEChannelImpl.cpp
```diff
@@ -0,0 +1,447 @@
...
+    // TODO: Confirm this generates a connection event and the AddConnection gets called so that the connection state is updated
+    return CHIP_NO_ERROR;
+}
+CHIP_ERROR BLEChannelImpl::SetConnectionParams(const Optional<uint8_t> & connectionHandle, uint32_t intervalMin,
+                                               uint32_t intervalMax, uint16_t latency, uint16_t timeout)
+{
+    sl_status_t ret;
+    if (connectionHandle.HasValue())
+    {
+        ret = sl_bt_connection_set_parameters(connectionHandle.Value(), intervalMin, intervalMax, latency, timeout, 0, 0xFFFF);
```
jepenven-silabs, line 353: Magic Numbers...

src/platform/silabs/efr32/BLEChannelImpl.cpp
```diff
@@ -0,0 +1,447 @@
...
+    sl_status_t ret = sl_bt_connection_open(address, addrType, sl_bt_gap_1m_phy, &mConnectionState.connectionHandle);
+    VerifyOrReturnError(ret == SL_STATUS_OK, MapBLEError(ret));
+
+    // TODO: Confirm this generates a connection event and the AddConnection gets called so that the connection state is updated
+    return CHIP_NO_ERROR;
+}
+CHIP_ERROR BLEChannelImpl::SetConnectionParams(const Optional<uint8_t> & connectionHandle, uint32_t intervalMin,
+                                               uint32_t intervalMax, uint16_t latency, uint16_t timeout)
+{
+    sl_status_t ret;
```
jepenven-silabs, line 350: default value missing

src/platform/silabs/efr32/BLEChannelImpl.cpp
```diff
@@ -0,0 +1,447 @@
...
+    mAdvMaxEvents   = maxEvents.ValueOr(0);
+
+    if (channelMap.HasValue())
+    {
+        ret = sl_bt_advertiser_set_channel_map(mAdvHandle, channelMap.Value());
+        VerifyOrReturnError(ret == SL_STATUS_OK, MapBLEError(ret));
+    }
+
+    ret = sl_bt_advertiser_set_timing(mAdvHandle, mAdvIntervalMin, mAdvIntervalMax, mAdvDuration, mAdvMaxEvents);
+    VerifyOrReturnError(ret == SL_STATUS_OK, MapBLEError(ret));
```
jepenven-silabs, line 381: Useless check, on peut juste faire un `return MapBLEError(ret);`

src/platform/silabs/efr32/BLEChannelImpl.cpp
```diff
@@ -0,0 +1,447 @@
...
+
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR BLEChannelImpl::CloseConnection()
+{
+    sl_status_t ret = sl_bt_connection_close(mConnectionState.connectionHandle);
+    VerifyOrReturnError(ret == SL_STATUS_OK, MapBLEError(ret));
```
jepenven-silabs, line 389: Useless check, on peut juste faire un return MapBLEError(ret);

src/platform/silabs/efr32/BLEChannelImpl.cpp
```diff
@@ -0,0 +1,447 @@
...
+    // Todo: Confirm this generates a disconnect event and the RemoveConnection gets called
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR BLEChannelImpl::SetAdvHandle(uint8_t handle)
+{
+    sl_status_t ret;
+    if (mAdvHandle == 0xff)
```
jepenven-silabs, line 397: ca pourrait pas etre plus optimiser du genre 
```suggestion
    if (handle == mAdvHandle) 
    {
        return CHIP_NO_ERROR;
    }
    
    sl_bt_advertiser_delete_set(mAdvHandle); // No need to check since we are creating a new set anyway
    mAdvHandle = handle;
    return MapBLEError(sl_bt_advertiser_create_set(&mAdvHandle));
    
```

Pas sur si c'est mieux au niveau behaviour par contre

src/platform/silabs/efr32/BLEChannelImpl.cpp
```diff
@@ -0,0 +1,447 @@
...
+
+        mAdvHandle = handle;
+        ret        = sl_bt_advertiser_create_set(&mAdvHandle);
+    }
+    return MapBLEError(ret);
+}
+
+CHIP_ERROR BLEChannelImpl::DiscoverServices()
+{
+    sl_status_t ret = sl_bt_gatt_discover_primary_services(mConnectionState.connectionHandle);
```
jepenven-silabs, line 419: Toutes ces fonctions pourraient etre des one liners 
```suggestion
    return MapBLEEroor(sl_bt_gatt_discover_primary_services(mConnectionState.connectionHandle));
```

src/platform/silabs/efr32/BUILD.gn
```diff
@@ -77,7 +77,12 @@ static_library("efr32") {
       "../rs911x/wfx_sl_ble_init.h",
     ]
   } else {
-    sources += [ "BLEManagerImpl.cpp" ]
+    sources += [
+      "BLEChannel.h",
```
jepenven-silabs, line 81: Fac de facto on va etre pris avec le BLE channel ? c'est pas un build option ? bonjour le code bloat... 

SVP rendre ca plus modulaire au niveau du build.

jepenven-silabs: En plus t'as rajouter un gn arg `sl_enable_ble_side_channel` faut juste l'appliquer ici! :)

jepenven-silabs (commented):
Quelques commentaires a fixer

src/platform/silabs/efr32/BLEChannelImpl.cpp
jepenven-silabs: My point still stands, Hardcoded arrays inside a CPP file should not be merged at all. There shouldn't be a need for that, so we need to move it elsewhere in the form of a compile time overiddable define like 

```
#ifndef SIDE_CHANNEL_UUID
#define SIDE_CHANNEL_UUID 0x01
#endif 
```

Because unless I'm missing something, it's looks like we are imposing debug/test/samples values to our customers without any easy way to modify them other than actually modifying a cpp file.

jepenven-silabs (approved):
Approving considering the Timeline

PR #443, opened 2025-05-08
[SL-ONLY] Fixing the build issue with the refrigerator app

chirag-silabs:
With the zap update all the clusters are having there own EventIds.h file. Hence Event::Notify::Id is available in RefrigeratorAlarm, RefrigeratorAndTemperatureControlledCabinetMode and TemperatureControl which is throwing the error 'Events' is ambiguous. 

Fixing it by specfying the Id which needs to be updated. 

#### Testing
Tested locally on 917 SoC shell command.


senthilku:
@chirag-silabs , Check the description part and update it.

mkardous-silabs (changes requested):
Refrigerator-app exists on CSA. The change should happen there.

chirag-silabs:
> Refrigerator-app exists on CSA. The change should happen there.

ohh my bad, though refrigerator was silabs only app, will create a PR on CSA

PR #446, opened 2025-05-09
[CSA-CP]Fix TestInetEndPoint to pass even when computer is offline (#38015)

jmartinez-silabs:
Description:

- TestInetEndPoint was not hermetic, it required at least one network interface connected to a link and up.
- There was a comment before to skip tests if the link was not up, but it was not actually skipping the test. This made it fail for me when working offline and trying to run unit tests.

Testing done:
- Test passes with and without link up.

#### Testing
Unit test ci


mkardous-silabs (approved):

jmartinez-silabs:
Tests / REPL Tests - Linux (pull_request is broken on CSA too, ill bypass merge this to unlock unit test on prs

PR #451, opened 2025-05-13
[CSA-CP] Adds fix for integer overflow in TicksToMS function

senthilku:
Cherry Pick of https://github.com/project-chip/connectedhomeip/pull/38757

#### Description
Adds fix for integer overflow in LWIP function which causes timer to not work post 70mins.

#### Testing
Manually tested using logs of LWIP timers with CSA PR


mkardous-silabs (approved):
I'm ok with adding the copilot comment but really doesn't add that much value in this case.

jmartinez-silabs (approved):

PR #463, opened 2025-05-20
[SL-UP] Fix/unused var in ble manager

jmartinez-silabs:
Build fails with unused variable error when apps are built without logs.
Fix: Remove the usage of the local variable

#### Testing
Build app with log disable/release


mkardous-silabs (approved):
Can we add a release build to the matter_sdk CI at the same time?

jepenven-silabs (approved):

PR #480, opened 2025-05-30
[SL-UP] Fixing the warning for the 917 SoC devices

chirag-silabs:
There were multiple warning flags enabled for the 917 SoC devices. Removing the warning flags

#### Testing
Tested with lighting and lock 917 SoC builds and CI

### DIFF
--- a/src/platform/silabs/ConfigurationManagerImpl.cpp
+++ b/src/platform/silabs/ConfigurationManagerImpl.cpp
@@ -91,9 +91,8 @@ CHIP_ERROR ConfigurationManagerImpl::GetBootReason(uint32_t & bootReason)
 {
     // rebootCause is obtained at bootup.
     BootReasonType matterBootCause;
-    uint32_t rebootCause = Silabs::GetPlatform().GetRebootCause();
-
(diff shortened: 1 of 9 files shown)


src/platform/silabs/ConfigurationManagerImpl.cpp
```diff
@@ -91,9 +91,8 @@ CHIP_ERROR ConfigurationManagerImpl::GetBootReason(uint32_t & bootReason)
 {
     // rebootCause is obtained at bootup.
     BootReasonType matterBootCause;
-    uint32_t rebootCause = Silabs::GetPlatform().GetRebootCause();
-
 #if defined(_RMU_RSTCAUSE_MASK)
+    uint32_t rebootCause = Silabs::GetPlatform().GetRebootCause();
```
jmartinez-silabs, line 95: My comment is the same as copilot. No need to duplicate the declaration.
Please put it back where it was and use
`[[maybe_unused]] uint32_t rebootCause = Silabs::GetPlatform().GetRebootCause();`

jmartinez-silabs (changes requested):

selissia (approved):

PR #488, opened 2025-06-04
[SL-TEMP] Adjust main task priority based on SLI_SI91X_MCU_INTERFACE configuration

rosahay-silabs:
### Issue
 - Priority inversion of the tasks happening because initialization task is getting blocked during the storage initialization causing the already lower priority BLE task to pick up pending events with clearing the event buffer but without clearing the flags from the event. Hence the higher priority task loops thinking that there is an event for it but since the clean-up is not done gracefully, it loops.
 - Temporary fix was asked from the application by elevating the command engine and event task priority, since the above mentioned scenario won't happen. The SDK was designed with this presumption that the command engine and event task are the highest priority task.
 - The fix from the WiFi SDK is planned for the 3.5.2 patch release with a longer test cycle.

### Changelog
 - Changing the priority of the `main` thread, responsible for the initialization of hardware.
 - Increase the priority of the command engine and event task of the WiFi SDK.
 - Changes restricted to 917 SoC only to reduce the impact of the change.

### Testing
Tested in the MTBF environment for over 16 hours without init failure.


senthilku:
Increase the priority of the command engine and event task of the WiFi SDK - This is updated in matter extension PR right?
https://stash.silabs.com/projects/WMN_TOOLS/repos/matter_extension/pull-requests/484/overview

senthilku (approved):

rosahay-silabs:
As discussed with @jmartinez-silabs , this is a SL-UP for now till we get the proper fix from the WiseConnect SDK.

PR #495, opened 2025-06-12
[SL-UP] BLEChannel added missing pure virtual

lpbeliveau-silabs:
#### Summary

Added the missing =0 in BLEChannel methods

#### Related issues

This would cause linking issue when building with optimization O0.

#### Testing

Built with O0, note the O0 code doesn't fit on our part so it couldn't be tested further.


jmartinez-silabs (commented):
need real title and Tag is [SL-UP] else script won't pickit up

PR #504, opened 2025-06-30
[CSA-CP]Fault Logging Fix

lpbeliveau-silabs:
#### Summary

Added a uart flush to force a uart output without relying on the uart task when we hit a fault, added handlers for faults other than hardfault to get logs.

#### Testing

Simulated faults with a wrongly addressed pointer by dereferencing it, observed logs would be generated.

Note: The equivalent PR is also up on CSA.

### DIFF
--- a/examples/platform/silabs/SoftwareFaultReports.cpp
+++ b/examples/platform/silabs/SoftwareFaultReports.cpp
@@ -147,7 +147,7 @@ extern "C" __attribute__((used)) void debugHardfault(uint32_t * sp)
  * Log a fault to the debugHardfault function.
  * This function is called by the fault handlers to log the fault details.
  */
-extern "C" void LogFault_Handler(void)
+extern "C" __attribute__((naked)) void LogFault_Handler(void)
 {
     uint32_t * sp;
     __asm volatile("tst lr, #4 \n"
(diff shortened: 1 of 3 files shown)


examples/platform/silabs/SoftwareFaultReports.cpp
```diff
@@ -131,24 +135,59 @@ extern "C" __attribute__((used)) void debugHardfault(uint32_t * sp)
...
+                   "mrseq %0, msp \n"
+                   "mrsne %0, psp \n"
+                   : "=r"(sp));
+    debugHardfault(sp);
+}
+
 #ifndef SL_CATALOG_ZIGBEE_STACK_COMMON_PRESENT
+extern "C" __attribute__((naked)) void NMI_Handler(void)
+{
+    __asm volatile("b LogFault_Handler");
```
jmartinez-silabs, line 164: why does it need to be an assembly branch operation? Why can we not just call ` LogFault_Handler();`

lpbeliveau-silabs: So my compiler knowledge is  bit rusty, but I think  that a C call would have the compiler would push the return address of LogFault_Handler on the callstack, which is not relevant debugging info. 

I don't think this would be dramatic either but I would have to test it cannot generate any other corruption of register/stack pointer with different optimization, so I felt it was safer to just have the assembly branch that doesn't alter anything else

lpbeliveau-silabs: I'm sure there are also proper guards that I could put around a C call to make this behave as I want as well, so I can also look into it if we really want to get rid of this assembly branch instruction

PR #511, opened 2025-07-14
[SL-UP] Adds support for PKCS7 padding in OTA flow

rosahay-silabs:
#### Summary

The payload of the application/firmware image may not be properly aligned, but for AES encryption the payload needs to be `block_size` aligned, in this case that is `16`. Currently, we are using a naive `0xFF` padding, but that assumes that the payload will not end with `0xFF` which is risky. Hence replacing the naive padding logic with PKCS7 standardized logic.

Since the padding is done on the file and not on the block, this means the last block of the file will have the padding data, hence the last block of the file needs to be identified and padding needs to be removed from that to get back the original file. 

### Testing
 - [x] Tested with 917 SoC
 - [x] Tested with 917 NCP, pending as SiSDK cloning is taking a lot of time.

### DIFF
--- a/scripts/tools/silabs/ota/crypto_utils.py
+++ b/scripts/tools/silabs/ota/crypto_utils.py
@@ -380,10 +380,11 @@ class rijndael:
 
 def encryptFlashData(nonce, key, data, imageLen):
     encyptedBlock = ''
-    if (imageLen % 16) != 0:
-        for x in range(16 - (imageLen % 16)):
-            data = data + bytes([255])
-        imageLen = len(data)
+    block_size = 16
+    # pad using PKCS#7 padding
+    pad_len = block_size - (imageLen % block_size)
+    data = data + bytes([pad_len] * pad_len)
+    imageLen = len(data)
 
     r = rijndael(key, block_size=16)
 
--- a/src/platform/silabs/multi-ota/OTAFirmwareProcessor.cpp
+++ b/src/platform/silabs/multi-ota/OTAFirmwareProcessor.cpp
@@ -72,6 +72,14 @@ CHIP_ERROR OTAFirmwareProcessor::ProcessInternal(ByteSpan & block)
     }
 
     OTATlvProcessor::vOtaProcessInternalEncryption(byteBlock);
+    if (IsLastBlock() == true)
+    {
+        // Unpad the file if it is the last block
+        // This is necessary to remove any padding that might have been added during encryption.
+        // The unpadding logic is specific to the encryption scheme used.
+        VerifyOrReturnError(OTATlvProcessor::UnpadFile(byteBlock) == CHIP_NO_ERROR, CHIP_ERROR_WRONG_ENCRYPTION_TYPE,
+                            ChipLogError(SoftwareUpdate, "UnpadFile failed"));
+    }
     block = byteBlock;
 #endif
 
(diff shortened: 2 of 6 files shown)


src/platform/silabs/multi-ota/SiWx917/OTAWiFiFirmwareProcessor.cpp
```diff
@@ -63,6 +63,14 @@ CHIP_ERROR OTAWiFiFirmwareProcessor::ProcessInternal(ByteSpan & block)
     }
 
     OTATlvProcessor::vOtaProcessInternalEncryption(byteBlock);
+    if (IsLastBlock() == true)
+    {
+        // Unpad the file if it is the last block
```
rcasallas-silabs, line 68: This documentation should be moved to either the method declaration (header), or the method definition (.cpp). Header is preferable.

src/platform/silabs/multi-ota/OTAFirmwareProcessor.cpp
```diff
@@ -72,6 +72,14 @@ CHIP_ERROR OTAFirmwareProcessor::ProcessInternal(ByteSpan & block)
     }
 
     OTATlvProcessor::vOtaProcessInternalEncryption(byteBlock);
+    if (IsLastBlock() == true)
+    {
+        // Unpad the file if it is the last block
+        // This is necessary to remove any padding that might have been added during encryption.
+        // The unpadding logic is specific to the encryption scheme used.
+        VerifyOrReturnError(OTATlvProcessor::UnpadFile(byteBlock) == CHIP_NO_ERROR, CHIP_ERROR_WRONG_ENCRYPTION_TYPE,
```
rcasallas-silabs, line 80:
```suggestion
        VerifyOrReturnError(OTATlvProcessor::RemovePadding(byteBlock) == CHIP_NO_ERROR, CHIP_ERROR_WRONG_ENCRYPTION_TYPE,
```

src/platform/silabs/multi-ota/OTAFirmwareProcessor.cpp
```diff
@@ -72,6 +72,14 @@ CHIP_ERROR OTAFirmwareProcessor::ProcessInternal(ByteSpan & block)
     }
 
     OTATlvProcessor::vOtaProcessInternalEncryption(byteBlock);
+    if (IsLastBlock() == true)
+    {
+        // Unpad the file if it is the last block
```
rcasallas-silabs, line 77: Move the documentation to the method declaration.

scripts/tools/silabs/ota/crypto_utils.py
```diff
@@ -380,10 +380,11 @@ def decrypt(self, ciphertext):
...
     encyptedBlock = ''
-    if (imageLen % 16) != 0:
-        for x in range(16 - (imageLen % 16)):
-            data = data + bytes([255])
-        imageLen = len(data)
+    block_size = 16
+    # pad using PKCS#7 padding
+    pad_len = block_size - (imageLen % block_size)
+    data = data + bytes([pad_len] * pad_len)
+    imageLen = len(data)
```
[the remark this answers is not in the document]
rosahay-silabs: `pad_len` will never be `0`, since `imageLen % block_size` will be in between 0 and 15, so `pad_len` will be `16` in case the payload is aligned to the `block_size`, as `imageLen % block_size` will be `0` in that case.

rcasallas-silabs (approved):

rcasallas-silabs (approved):

rosahay-silabs:
> This affects Thread too, right? So the Thread case also needs to be tested.

Yes, I was unable to test due to SiSDK taking much longer. I will test it (NCP) today before merging, and since codepath is same for both I am pretty sure it should work.

PR #536, opened 2025-08-05
[SL-ONLY] Fixing google nest sleep with iphone

chirag-silabs:
#### Summary
When google ecosystem is commissioned with google home app running on the Iphone/Ipad, the apple product creates a Key Chain pair which it does with the Apple ecosystem. This fabric is not subscribed causing the device to go only in the DTIM sleep not the LI sleep. 

Adding the condition to let the device go to LI sleep when the google home is commissioned with the apple device.

#### Related issues
https://jira.silabs.com/browse/MATTER-5273

#### Testing
Tested on the 917SoC with google home, apple and amazon device.


senthilku (approved):

examples/platform/silabs/wifi/icd/ApplicationSleepManager.cpp
```diff
@@ -152,7 +152,7 @@ bool ApplicationSleepManager::ProcessKeychainEdgeCase()
 
     for (auto it = mFabricTable->begin(); it != mFabricTable->end(); ++it)
     {
-        if ((it->GetVendorId() == chip::VendorId::Apple) &&
+        if ((it->GetVendorId() == chip::VendorId::Apple || it->GetVendorId() == chip::VendorId::Google) &&
```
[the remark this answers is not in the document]
chirag-silabs: The function name tells about the processing the keychain which is common. Hence resolving the conversation.

selissia, line 155: Does the function description need to be updated in ApplicationSleepManager.h ?

selissia (approved):

jepenven-silabs (approved):

examples/platform/silabs/wifi/icd/ApplicationSleepManager.cpp
chirag-silabs: Updated the function description

mkardous-silabs:
@chirag-silabs This is already fixed on matter_sdk::main. We should use the same fix instead of adding the if condition.
Based on the if, if i am not mistaken, we are risking adding unexpected behaviors for the google fabric.

mkardous-silabs:
@chirag-silabs Renamed the PR to SL-TEMP even if the commit name won't change since this doesn't need to be upstreamed to main.

PR #565, opened 2025-08-20
[CSA-CP] Adding sl_main support for the 917SoC

chirag-silabs:
#### Summary
The sl_main support wasn't added to the 917SoC on the release branch

Adding the sl_main support to the 917SoC

#### Testing
Tested locally with 917SoC-


chirag-silabs:
PR no longer needed.

PR #567, opened 2025-08-22
Fixes Matter AWS connection stability issues

sabollim-silabs:
#### Summary
This Pull request aims to fix the intermettient connectivity failures seen on MatterAws componet.

the fixes dones as part of it are:
1) Changed MATTER_AWS task priority to osPriorityAboveNormal ( to be in syns with cmsis_o2 priorities)
2) Adopted the general priority used for TCPIP thread (previously the thread priority for MatterAws is diferent)
3) Added few Macro define guards to avoid redifinition failures in slcp build
4) Added altcp_output after altcp_write in mbedtls_bio_send to make sure the packet is send after write , previously the issue was the packets are being queued in ip layer and the packet len and payload was not alligned for queued packets.
this causes failure during mbedtls_ssl_handshake.

#### Testing
build with MatterAws compnent enabled and test AWS connectivity(10 iterations)


src/lwip/silabs/lwipopts-rs911x.h
```diff
@@ -67,18 +67,17 @@
 #define TCP_QUEUE_OOSEQ 0
 #define ARP_QUEUEING (0)
 #define TCPIP_THREAD_NAME "LWIP"
+#define TCPIP_THREAD_PRIO osPriorityAboveNormal
```
jmartinez-silabs, line 70: How was this decided? It is neither `osPriorityRealtime7` or `osPriorityAboveNormal` like it was before

sabollim-silabs: osPriorityAboveNormal is already used priority for TCPIP thread, previously DIC used a different priority due to stability issues. now i removed as we fixed the stability issues and using the osPriorityAboveNormal

jmartinez-silabs: My mistake, must have read too fast. I was under the impression that the new Prio was different than before

PR #574, opened 2025-08-26
[SL-ONLY]Use existing string for tracing class to validate shell cmd

jmartinez-silabs:
#### Summary
Small optimization to reuse the string defined in SilabsTracing.cpp for all TimeTraceOperation Enums values in the tracing shell commands arguments.

This makes it so you can add a new enum/string entry in 1 location SilabsTracing.cpp and make it available in shell commands without duplicating the string.

#### Related issues
N/A

#### Testing
Validate that the Shell tracing command can recognize all/any `TimeTraceOperation` string as defined in `TimeTraceOperationToString` from third_party/matter_sdk/src/platform/silabs/tracing/SilabsTracing.cpp

### DIFF
--- a/examples/platform/silabs/shell/tracing/TracingShellCommands.cpp
+++ b/examples/platform/silabs/shell/tracing/TracingShellCommands.cpp
@@ -34,11 +34,12 @@ using SilabsTracer       = Tracing::Silabs::SilabsTracer;
 
 TimeTraceOperation StringToTimeTraceOperation(const char * str)
 {
-    for (uint32_t ttOp = 0; ttOp < to_underlying(TimeTraceOperation::kNumTraces); ttOp++)
+    for (auto ttOp = 0; ttOp < to_underlying(TimeTraceOperation::kNumTraces); ttOp++)
     {
-        if (strcmp(str, TimeTraceOperationToString(static_cast<TimeTraceOperation>(ttOp))) == 0)
+        TimeTraceOperation op = static_cast<TimeTraceOperation>(ttOp);
+        if (strcmp(str, TimeTraceOperationToString(op)) == 0)
         {
-            return static_cast<TimeTraceOperation>(ttOp);
+            return op;
         }
     }
     return TimeTraceOperation::kNumTraces;


examples/platform/silabs/shell/tracing/TracingShellCommands.cpp
```diff
@@ -34,102 +34,14 @@ using SilabsTracer       = Tracing::Silabs::SilabsTracer;
...
-        return TimeTraceOperation::kMatterInit;
-    }
-    else if (strcmp(str, "BufferFull") == 0)
-    {
-        return TimeTraceOperation::kBufferFull;
-    }
-    else
-    {
-        return TimeTraceOperation::kNumTraces;
+        if (strcmp(str, TimeTraceOperationToString(static_cast<TimeTraceOperation>(ttOp))) == 0)
```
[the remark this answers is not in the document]
jmartinez-silabs: Can't cache the value, as it will change for each iteration. The call leads to a switch case to execute the lookup, no expensive operations are done.

examples/platform/silabs/shell/tracing/TracingShellCommands.cpp
```diff
@@ -34,102 +34,14 @@ using SilabsTracer       = Tracing::Silabs::SilabsTracer;
 
 TimeTraceOperation StringToTimeTraceOperation(const char * str)
 {
-    if (strcmp(str, "Spake2p") == 0)
+    for (uint32_t ttOp = 0; ttOp < to_underlying(TimeTraceOperation::kNumTraces); ttOp++)
```
mkardous-silabs, line 37: Enum type is an uint8 not a uint32.
```suggestion
    for (auto ttOp = 0; ttOp < to_underlying(TimeTraceOperation::kNumTraces); ttOp++)
```

examples/platform/silabs/shell/tracing/TracingShellCommands.cpp
mkardous-silabs, line 39: To reduce the number of casts, you coul d

op = static_cast... and reuse the same variable afterwards.

mkardous-silabs (approved):

examples/platform/silabs/shell/tracing/TracingShellCommands.cpp
jmartinez-silabs: done

examples/platform/silabs/shell/tracing/TracingShellCommands.cpp
jmartinez-silabs: done

PR #603, opened 2025-09-12
[SL-UP] Migrate matter certificates on Series3 to a different location

jmartinez-silabs:
#### Summary
Issue:
Zigbee started using the static token from the Common Token Manager and is storing a token on top of our certificates, corrupting them. We cannot use static tokens in Matter for our certificates due to the lack of an API to replace them if needed.

Solution: 
2 pages of 4k (total 8k) are reserved for all tokens. Currently, there isn't a lot of static token written by the Zigbee stack. So we move our certificate to the last page for the time being. This is a workaround until CTM can be extended and used for Matter Cert.

However, the corruption occurs during the first run of the Zigbee stack task, which is of high priority. With sl_main, the scheduler is started before our app init. The corruption occurs before our migration manager can run.  Some inits and the Migration Manager had to be moved to an earlier time, and it had to make sure other tasks didn't interrupt the migration.

The issue only affects Series 3 and technically only CMP app (which runs Zibee also).

A change that is not related to the issue, but is needed due to the changes are changes to the older migration of KVS table. Since migration now runs on sl_main, stack usage was an issue. I preferred using Heap allocation, which is then freed, than increasing the stack size of the sl_main task for all apps.

#### Related issues
https://jira.silabs.com/browse/MATTER-5455

#### Testing
Provision Series3,
Validate that migration is done and certificates are moved and intact with commander readmem
Commission the device successfully using the provisioned certificates.

Tested that 917 can still boot and run the migration manager if need be, and commission it to validate it still works.

### DIFF
--- a/src/platform/silabs/KeyValueStoreManagerImpl.cpp
+++ b/src/platform/silabs/KeyValueStoreManagerImpl.cpp
@@ -24,6 +24,7 @@
 #include "MigrationManager.h"
 #include <cmsis_os2.h>
 #include <crypto/CHIPCryptoPAL.h>
+#include <lib/support/ScopedBuffer.h>
 #include <platform/CHIPDeviceLayer.h>
 #include <platform/KeyValueStoreManager.h>
 #include <platform/silabs/SilabsConfig.h>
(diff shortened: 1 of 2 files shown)


src/platform/silabs/platformAbstraction/BUILD.gn
```diff
@@ -24,6 +24,8 @@ source_set("common") {
     "${chip_root}/src/platform/silabs/platformAbstraction/SilabsPlatform.h",
     "${chip_root}/src/platform/silabs/platformAbstraction/SilabsPlatformBase.h",
   ]
+
+  include_dirs = [ "${chip_root}/src/platform/silabs" ]
```
mkardous-silabs, line 28:
```suggestion
  include_dirs = [ "${chip_root}/src/platform/silabs" ]
  
```

src/platform/silabs/MigrationManager.cpp
```diff
@@ -132,6 +170,73 @@ void MigrateHardwareVersion(void)
...
+            uint8_t * cdBuffer  = new uint8_t[cdSize];
+            actionSucceed       = (dacBuffer != nullptr && paiBuffer != nullptr && cdBuffer != nullptr);
+            MutableByteSpan dacBufferSpan(dacBuffer, dacSize);
+            MutableByteSpan paiBufferSpan(paiBuffer, paiSize);
+            MutableByteSpan cdBufferSpan(cdBuffer, cdSize);
+
+            // Step to read the certificates at their current location
+            if (actionSucceed)
+            {
+                provision.Init();
```
mkardous-silabs, line 212: This causes a double provision init for Wi-Fi apps. Not sure if it is the case other flavors.

src/platform/silabs/MigrationManager.cpp
```diff
@@ -132,6 +170,73 @@ void MigrateHardwareVersion(void)
...
+                actionSucceed &= (provision.GetStorage().GetCertificationDeclaration(cdBufferSpan) == CHIP_NO_ERROR);
+            }
+
+            // Step to write the certificates to their new location
+            if (actionSucceed)
+            {
+                provision.GetStorage().Initialize(0, 0);
+                provision.GetStorage().SetCredentialsBaseAddress(secondPageAddr);
+                // Write all certs back to the second page
+                // On the first write, we erase the new page.
```
mkardous-silabs, line 225: where is the erase done?

src/platform/silabs/KeyValueStoreManagerImpl.cpp
```diff
@@ -275,33 +271,46 @@ void KeyValueStoreManagerImpl::ErasePartition(void)
...
+    size_t readlen                   = 0;
+    constexpr uint32_t oldMaxEntries = 120;
+    uint32_t maxStringLen            = PersistentStorageDelegate::kKeyLengthMax + 1;
+    char * mKvsStoredKeyString       = new char[oldMaxEntries * maxStringLen];
+
+    CHIP_ERROR err;
+    if (!mKvsStoredKeyString)
+    {
+        err = CHIP_ERROR_NO_MEMORY;
+    }
```
mkardous-silabs, line 285: This could be replaced with a `VerifyOrReturn` to simplify the code execution.

mkardous-silabs (approved):

src/platform/silabs/MigrationManager.cpp
jmartinez-silabs: In the function under the comment on the first write after Initialize

PR #611, opened 2025-09-16
[SL-UP] Reset retry interval on WiFi connection notification

rosahay-silabs:
#### Summary

The retry timer is not getting reset on a successful connection after a retry, thus consecutive retry attempts are set as per the existing timer value.

#### Related issues

NOJIRA

#### Testing

This was tested by rebooting the access point.

#### Readability checklist

code readable:

-   [x] PR title is
-   [ ] Apply the
        rule (coding style)
-   [ ] PR size is short

See: [Pull Request Guidelines](https://project-chip.github.io/connectedhomeip-doc/contributing/pull_request_guidelines.html)

### DIFF
--- a/src/platform/silabs/NetworkCommissioningWiFiDriver.cpp
+++ b/src/platform/silabs/NetworkCommissioningWiFiDriver.cpp
@@ -163,6 +163,8 @@ CHIP_ERROR SlWiFiDriver::ConnectWiFiNetwork(const char * ssid, uint8_t ssidLen, 
     wifiConfig.security = WFX_SEC_WPA2;
 
     ChipLogProgress(NetworkProvisioning, "Setting up connection for WiFi SSID: %.*s", static_cast<int>(ssidLen), ssid);
+    // Resetting the retry connection state machine for a new access point connection
+    WifiInterface::GetInstance().ResetScheduledConnectionAttempts();
     // Configure the WFX WiFi interface.
     WifiInterface::GetInstance().SetWifiCredentials(wifiConfig);
     ReturnErrorOnFailure(ConnectivityMgr().SetWiFiStationMode(ConnectivityManager::kWiFiStationMode_Disabled));
--- a/src/platform/silabs/NetworkCommissioningWiFiDriver.cpp
+++ b/src/platform/silabs/NetworkCommissioningWiFiDriver.cpp
@@ -164,7 +164,7 @@ CHIP_ERROR SlWiFiDriver::ConnectWiFiNetwork(const char * ssid, uint8_t ssidLen, 
 
     ChipLogProgress(NetworkProvisioning, "Setting up connection for WiFi SSID: %.*s", static_cast<int>(ssidLen), ssid);
     // Resetting the retry connection state machine for a new access point connection
-    WifiInterface::GetInstance().ResetScheduledConnectionAttempts();
+    WifiInterface::GetInstance().ResetConnectionRetryInterval();
     // Configure the WFX WiFi interface.
     WifiInterface::GetInstance().SetWifiCredentials(wifiConfig);
     ReturnErrorOnFailure(ConnectivityMgr().SetWiFiStationMode(ConnectivityManager::kWiFiStationMode_Disabled));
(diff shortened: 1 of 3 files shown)


src/platform/silabs/wifi/WifiInterface.cpp
```diff
@@ -106,6 +106,7 @@ void WifiInterface::NotifyConnection(const MacAddress & ap)
     evt.body.channel = wfx_rsi.ap_chan;
 #endif
     std::copy(ap.begin(), ap.end(), evt.body.mac);
+    retryInterval = kWlanMinRetryIntervalsInSec;
```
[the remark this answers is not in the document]
rosahay-silabs: @jmartinez-silabs , yes that makes total sense. Let me check if we can implement it at a suitable place or if we need some API to do that.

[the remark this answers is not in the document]
rosahay-silabs: hey @jmartinez-silabs we still have to keep this here, the reason being, lets say during the first connection attempt the access point connected after 4-5 attempts, after which lets say after 5-10mins the access point gets disconnected. Now for this retry attempt the timer should reset and should not resume from the value from the previous attempts.

jmartinez-silabs (approved):

src/platform/silabs/wifi/WifiInterface.cpp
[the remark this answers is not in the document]
rosahay-silabs: The retry timer is initialized in ConnectWifiNetwork to handle new connection attempts and SSID changes. On successful connection in JoinWifiNetwork, the timer is reset to prepare for future retries after disconnection. We avoid resetting the timer in InitiateScan to ensure consistent retry behavior, since retries loop through InitiateScan and JoinWifiNetwork without restarting from ConnectWifiNetwork.

src/platform/silabs/wifi/WifiInterface.cpp
```diff
@@ -176,6 +179,12 @@ void WifiInterface::ScheduleConnectionAttempt()
     retryInterval += retryInterval;
 }
 
+void WifiInterface::ResetScheduledConnectionAttempts()
```
mkardous-silabs, line 182:
```suggestion
void WifiInterface::ResetConnectionRetryInterval()
```

src/platform/silabs/wifi/WifiInterface.h
```diff
@@ -353,6 +353,11 @@ class WifiInterface : public WifiStateProvider, public PowerSaveInterface
         return static_cast<uint32_t>(1UL << chip::to_underlying(chip::app::Clusters::NetworkCommissioning::WiFiBandEnum::k2g4));
     }
 
+    /**
+     * @brief Function will reset the state of the reconnection attempts with the Access Point
```
mkardous-silabs, line 357:
```suggestion
     * @brief Function resets reconnection attempt interval back to the minimum value
```

src/platform/silabs/NetworkCommissioningWiFiDriver.cpp
```diff
@@ -163,6 +163,8 @@ CHIP_ERROR SlWiFiDriver::ConnectWiFiNetwork(const char * ssid, uint8_t ssidLen,
     wifiConfig.security = WFX_SEC_WPA2;
 
     ChipLogProgress(NetworkProvisioning, "Setting up connection for WiFi SSID: %.*s", static_cast<int>(ssidLen), ssid);
+    // Resetting the retry connection state machine for a new access point connection
+    WifiInterface::GetInstance().ResetScheduledConnectionAttempts();
```
mkardous-silabs, line 167: Why do we need the reset call here and in the WifiInterface?

rosahay-silabs: > What if the connection never succeeds. And later on there is a new connection attempt to a new ssid?
> Looking at the code, it looks like that new connection attempt will also start at whatever the last retryInterval was at.
>
> I think we should do retryInterval = kWlanMinRetryIntervalsInSec; at the start of a new connection attempt and then you don't need to worry about resetting it on a succeeded or failed connection

It is to cover this use case, though currently it is not coming up this way.

PR #631, opened 2025-09-18
[SL-UP] Increase KvsKeyMapCleanup stack to address overflow risk

jmartinez-silabs:
#### Summary
On Series 3, nvm3 secure API can sometimes use up to 1100 bytes on the stack.
`KvsKeyMapCleanupTask` is a task that runs once at init and can execute multiple nvm3 read or write successively.
In a rare occurrence, on power cycling of the device, a Stack overflow of this task occurred.

This PR increases the KvsKeyMapCleanupTask stack size to 1536 bytes, a 1024-byte increase. While this is targeted at series 3 use case, we can safely apply it to all our platforms and avoid conditional/configuration logic.
The stack of the task is dynamically allocated, and the task is deleted once the cleanup is completed, freeing the 1536 bytes allocated.

#### Related issues
MATTER-5377
MATTER-5397

#### Testing
Tested by building and running the init sequence on Series 2, Series 3 and 917 soc and verifying the task stack usage watermarks before task deletion

### DIFF
--- a/src/platform/silabs/KeyValueStoreManagerImpl.cpp
+++ b/src/platform/silabs/KeyValueStoreManagerImpl.cpp
@@ -62,7 +62,10 @@ CHIP_ERROR KeyValueStoreManagerImpl::Init(void)
     }
     ReturnErrorOnFailure(error);
 
-    constexpr osThreadAttr_t attr = { .name = "KvsKeyMapCleanupTask", .stack_size = 512 /* bytes */, .priority = osPriorityLow7 };
+    // On series 3, nvm3 security can use up to ~1100 bytes of stack during read/write operations.
+    // Provide enough stack for all platforms. The KvsKeyMapCleanupTask is deleted once the cleanup is completed and the allocated
+    // stack is freed.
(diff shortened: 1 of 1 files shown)


rosahay-silabs (approved):
Just a question, I understand this allocation will be temporary, but how are we sure, that there is no known scenario during this clean up task that will take up the additional 1024 bytes ?

jmartinez-silabs:
@rosahay-silabs you mean in the future or right now?
RIght now the case is specific to NVM3 security API, that only exists in Series3. The platform/nvm3 team confirmed that NVM3 crypto api can take up to 1100 bytes. Other than that, there is no other operation in the task than a for loop and nvm3 read nvm3 writes. So the stack usage is understood with the current version. The bump here is needed since in past sisdk nvm3 crypto api for Series3 didn't use that sl_crypto gcm api. 


If that is not your question then I am not sure i understand we can sync offline

jmartinez-silabs:
I have read the CLA Document and I hereby sign the CLA on behalf of myself and Silicon Labs, e-mail: junior.martinez@silabs.com

PR #929, opened 2026-04-22
[SL-ONLY] Buffering the packets until device is connected

chirag-silabs:
#### Summary
For the LIT device we are trying to disconnect from the AP and reconnect when we want to send any attribute update or while reporting data of subscription.
For the reporting data, that will be on the idle mode duration we can connect a bit early i.e., a part of different PR. but while posting attribute change the matter stack sends the packet unaware about the low layer transport is offline. 

To fix this, we are buffering the packets on the LwIP layer and flushing them once we are reconnected.
This have CSA code modification of the UDPEndpointLwIPImpl.cpp and .h files but it won't affect/conflicts in the future since the changes are moved a different file all together.

Unit test is added for the Queue impl to ensure we don't see any issues

#### Related issues
NA

#### Testing
Tested with 917SoC

### DIFF
--- a/src/inet/BUILD.gn
+++ b/src/inet/BUILD.gn
@@ -50,6 +50,10 @@ buildconfig_header("inet_buildconfig") {
     "INET_CONFIG_ENABLE_TCP_ENDPOINT=${chip_inet_config_enable_tcp_endpoint}",
     "INET_CONFIG_ENABLE_UDP_ENDPOINT=${chip_inet_config_enable_udp_endpoint}",
     "HAVE_LWIP_RAW_BIND_NETIF=true",
+
+    # [SL-ONLY] Queuing the UDP packets for till the netif is not up
+    "SILABS_INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY=${silabs_inet_config_udp_lwip_queue_until_netif_ready}",
+    "SILABS_INET_CONFIG_UDP_LWIP_DEFERRED_SEND_QUEUE_SIZE=${silabs_inet_config_udp_lwip_deferred_send_queue_size}",
   ]
 
   if (chip_inet_project_config_include != "") {
--- a/src/inet/BUILD.gn
+++ b/src/inet/BUILD.gn
@@ -51,7 +51,7 @@ buildconfig_header("inet_buildconfig") {
     "INET_CONFIG_ENABLE_UDP_ENDPOINT=${chip_inet_config_enable_udp_endpoint}",
     "HAVE_LWIP_RAW_BIND_NETIF=true",
 
-    # [SL-ONLY] Queuing the UDP packets for till the netif is not up
+    # [SL-ONLY] Queuing the UDP packets for until the netif is up
     "SILABS_INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY=${silabs_inet_config_udp_lwip_queue_until_netif_ready}",
     "SILABS_INET_CONFIG_UDP_LWIP_DEFERRED_SEND_QUEUE_SIZE=${silabs_inet_config_udp_lwip_deferred_send_queue_size}",
   ]
--- a/src/inet/BUILD.gn
+++ b/src/inet/BUILD.gn
@@ -193,6 +193,12 @@ static_library("inet") {
 
     if (!chip_system_config_use_network_framework) {
       sources += [ "UDPEndPointImpl${chip_system_config_inet}.cpp" ]
+      if (chip_system_config_inet == "LwIP") {
+        sources += [
+          "DeferredUdpSendQueueLwIP.cpp",
+          "DeferredUdpSendQueueLwIP.h",
+        ]
+      }
     }
   }
 
--- a/src/inet/BUILD.gn
+++ b/src/inet/BUILD.gn
@@ -52,8 +52,8 @@ buildconfig_header("inet_buildconfig") {
     "HAVE_LWIP_RAW_BIND_NETIF=true",
 
     # [SL-ONLY] Queuing the UDP packets for until the netif is up
-    "SILABS_INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY=${silabs_inet_config_udp_lwip_queue_until_netif_ready}",
-    "SILABS_INET_CONFIG_UDP_LWIP_DEFERRED_SEND_QUEUE_SIZE=${silabs_inet_config_udp_lwip_deferred_send_queue_size}",
+    "SL_INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY=${sl_inet_config_udp_lwip_queue_until_netif_ready}",
+    "SL_INET_CONFIG_UDP_LWIP_DEFERRED_SEND_QUEUE_SIZE=${sl_inet_config_udp_lwip_deferred_send_queue_size}",
   ]
 
   if (chip_inet_project_config_include != "") {
--- a/src/inet/BUILD.gn
+++ b/src/inet/BUILD.gn
@@ -187,6 +187,7 @@ static_library("inet") {
     #  - UDPEndPointImplSockets.cpp
     #  - UDPEndPointImplSockets.h
     sources += [
+      "DeferredUdpSendRing.h",
       "BasicPacketFilters.h",
       "EndpointQueueFilter.h",
       "UDPEndPoint.cpp",
--- a/src/inet/BUILD.gn
+++ b/src/inet/BUILD.gn
@@ -187,8 +187,8 @@ static_library("inet") {
     #  - UDPEndPointImplSockets.cpp
     #  - UDPEndPointImplSockets.h
     sources += [
-      "DeferredUdpSendRing.h",
       "BasicPacketFilters.h",
+      "DeferredUdpSendRing.h",
       "EndpointQueueFilter.h",
       "UDPEndPoint.cpp",
       "UDPEndPoint.h",
--- a/src/inet/BUILD.gn
+++ b/src/inet/BUILD.gn
@@ -198,7 +198,7 @@ static_library("inet") {
 
     if (!chip_system_config_use_network_framework) {
       sources += [ "UDPEndPointImpl${chip_system_config_inet}.cpp" ]
-      if (chip_system_config_inet == "LwIP") {
+      if (sl_inet_config_udp_lwip_queue_until_netif_ready) {
         sources += [
           "DeferredUdpSendQueueLwIP.cpp",
           "DeferredUdpSendQueueLwIP.h",
--- a/src/inet/BUILD.gn
+++ b/src/inet/BUILD.gn
@@ -51,7 +51,7 @@ buildconfig_header("inet_buildconfig") {
     "INET_CONFIG_ENABLE_UDP_ENDPOINT=${chip_inet_config_enable_udp_endpoint}",
     "HAVE_LWIP_RAW_BIND_NETIF=true",
 
-    # [SL-ONLY] Queuing the UDP packets for until the netif is up
+    # [SL-ONLY] Queuing the UDP packets until the netif is up
     "SL_INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY=${sl_inet_config_udp_lwip_queue_until_netif_ready}",
     "SL_INET_CONFIG_UDP_LWIP_DEFERRED_SEND_QUEUE_SIZE=${sl_inet_config_udp_lwip_deferred_send_queue_size}",
   ]
--- a/src/inet/DeferredUdpSendQueueLwIP.cpp
+++ b/src/inet/DeferredUdpSendQueueLwIP.cpp
@@ -19,6 +19,7 @@
 
 #if SL_INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY
 
+#include <inet/DeferredUdpSendRing.h>
 #include <inet/EndPointStateLwIP.h>
 #include <inet/IPAddress.h>
 #include <inet/IPPacketInfo.h>
@@ -26,7 +27,6 @@
 #include <inet/UDPEndPointImplLwIP.h>
 
 #include <lib/support/CodeUtils.h>
-#include <lib/support/FixedBuffer.h>
 #include <lib/support/logging/CHIPLogging.h>
 
 #include <lwip/err.h>
@@ -56,48 +56,7 @@ struct DeferredUdpSlot
     System::PacketBufferHandle msg;
 };
 
-struct DeferredUdpRing
-{
-    chip::FixedBuffer<DeferredUdpSlot, kDeferredQueueCapacity> slots{};
-    size_t head  = 0;
-    size_t count = 0;
-
-    bool Empty() const { return count == 0; }
-
-    void PushBack(DeferredUdpSlot && slot)
-    {
-        if (count == kDeferredQueueCapacity)
-        {
-            ChipLogDetail(Inet, "Deferred UDP queue full: dropping head");
-            slots[head] = DeferredUdpSlot{};
-            head        = (head + 1) % kDeferredQueueCapacity;
-            --count;
-        }
-        const size_t idx = (head + count) % kDeferredQueueCapacity;
-        slots[idx]       = std::move(slot);
-        ++count;
-    }
-
-    DeferredUdpSlot PopFront()
-    {
-        DeferredUdpSlot out = std::move(slots[head]);
-        slots[head]         = DeferredUdpSlot{};
-        head                = (head + 1) % kDeferredQueueCapacity;
-        --count;
-        return out;
-    }
-};
-
-/** Swap ring state (head, count) and per-slot storage without copying whole FixedBuffer (avoids copy-only path). */
-void SwapDeferredRings(DeferredUdpRing & a, DeferredUdpRing & b)
-{
-    std::swap(a.head, b.head);
-    std::swap(a.count, b.count);
-    for (size_t i = 0; i < kDeferredQueueCapacity; ++i)
-    {
-        std::swap(a.slots[i], b.slots[i]);
-    }
-}
+using DeferredUdpRing = DeferredUdpSendRing<DeferredUdpSlot, kDeferredQueueCapacity>;
 
 DeferredUdpRing gDeferredRing;
 
@@ -184,7 +143,10 @@ CHIP_ERROR DeferredUdpSendQueueLwIP::Enqueue(UDPEndPointImplLwIP * self, const I
     slot.msg      = std::move(msg);
 
     ChipLogDetail(Inet, "UDP send deferred (waiting for netif)");
-
+    if (gDeferredRing.IsFull())
+    {
+        ChipLogDetail(Inet, "Deferred UDP queue full: dropping head");
+    }
     gDeferredRing.PushBack(std::move(slot));
     return CHIP_NO_ERROR;
 }
@@ -192,7 +154,7 @@ CHIP_ERROR DeferredUdpSendQueueLwIP::Enqueue(UDPEndPointImplLwIP * self, const I
 void DeferredUdpSendQueueLwIP::PurgeForEndpoint(UDPEndPointImplLwIP * self)
 {
     UDPEndPoint * asBase = static_cast<UDPEndPoint *>(self);
-    const size_t n       = gDeferredRing.count;
+    const size_t n       = gDeferredRing.ActiveCount();
     for (size_t i = 0; i < n; ++i)
     {
         DeferredUdpSlot slot = gDeferredRing.PopFront();
@@ -209,7 +171,7 @@ void DeferredUdpSendQueueLwIP::Flush()
     assertChipStackLockedByCurrentThread();
 
     DeferredUdpRing pending{};
-    SwapDeferredRings(gDeferredRing, pending);
+    SwapDeferredUdpSendRings(gDeferredRing, pending);
 
     while (!pending.Empty())
     {
--- a/src/inet/DeferredUdpSendQueueLwIP.cpp
+++ b/src/inet/DeferredUdpSendQueueLwIP.cpp
@@ -111,7 +111,6 @@ bool IsNetifUsable(struct netif * netif)
     return netif != nullptr && netif_is_up(netif) && netif_is_link_up(netif);
 }
 
-#if LWIP_IPV6
 bool NetifHasValidIpv6Address(struct netif * netif)
 {
     VerifyOrReturnValue(netif != nullptr, false);
@@ -124,22 +123,14 @@ bool NetifHasValidIpv6Address(struct netif * netif)
     }
     return false;
 }
-#endif
 
 bool IsNetifReadyForOutboundUdp(struct netif * netif, const IPAddress & dest)
 {
     VerifyOrReturnValue(IsNetifUsable(netif), false);
-#if LWIP_IPV6
     if (dest.IsIPv6())
     {
         return NetifHasValidIpv6Address(netif);
     }
-#else
-    if (dest.IsIPv6())
-    {
-        return false;
-    }
-#endif
 #if INET_CONFIG_ENABLE_IPV4 && LWIP_IPV4
     if (dest.IsIPv4())
     {
--- a/src/inet/DeferredUdpSendQueueLwIP.cpp
+++ b/src/inet/DeferredUdpSendQueueLwIP.cpp
@@ -1,0 +1,267 @@
+/*
+ *
+ *    Copyright (c) 2020-2025 Project CHIP Authors
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+#include <inet/DeferredUdpSendQueueLwIP.h>
+
+#if INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY
+
+#include <inet/EndPointStateLwIP.h>
+#include <inet/IPAddress.h>
+#include <inet/IPPacketInfo.h>
+#include <inet/UDPEndPoint.h>
+#include <inet/UDPEndPointImplLwIP.h>
+
+#include <lib/support/CodeUtils.h>
+#include <lib/support/FixedBuffer.h>
+#include <lib/support/logging/CHIPLogging.h>
+
+#include <lwip/err.h>
+#include <lwip/ip.h>
+#include <lwip/netif.h>
+
+#include <platform/LockTracker.h>
+
+#include <utility>
+
+namespace chip {
+namespace Inet {
+
+namespace {
+
+#ifndef NETIF_FOREACH
+#define NETIF_FOREACH(netif) for ((netif) = netif_list; (netif) != nullptr; (netif) = (netif)->next)
+#endif
+
+constexpr size_t kDeferredQueueCapacity = INET_CONFIG_UDP_LWIP_DEFERRED_SEND_QUEUE_SIZE;
+static_assert(kDeferredQueueCapacity > 0, "INET_CONFIG_UDP_LWIP_DEFERRED_SEND_QUEUE_SIZE must be > 0");
+
+struct DeferredUdpSlot
+{
+    UDPEndPointHandle epHandle;
+    IPPacketInfo pktInfo;
+    System::PacketBufferHandle msg;
+};
+
+struct DeferredUdpRing
+{
+    chip::FixedBuffer<DeferredUdpSlot, kDeferredQueueCapacity> slots{};
+    size_t head  = 0;
+    size_t count = 0;
+
+    bool Empty() const { return count == 0; }
+
+    void PushBack(DeferredUdpSlot && slot)
+    {
+        if (count == kDeferredQueueCapacity)
+        {
+            const DeferredUdpSlot & dropped = slots[head];
+            char dropDest[IPAddress::kMaxStringLength];
+            dropped.pktInfo.DestAddress.ToString(dropDest);
+            ChipLogProgress(Inet, "Deferred UDP queue full: dropping head dest %s port %u len %u", dropDest, dropped.pktInfo.DestPort,
+                            static_cast<unsigned>(dropped.msg->TotalLength()));
+            slots[head] = DeferredUdpSlot{};
+            head        = (head + 1) % kDeferredQueueCapacity;
+            --count;
+        }
+        const size_t idx = (head + count) % kDeferredQueueCapacity;
+        slots[idx]       = std::move(slot);
+        ++count;
+    }
+
+    DeferredUdpSlot PopFront()
+    {
+        DeferredUdpSlot out = std::move(slots[head]);
+        slots[head]         = DeferredUdpSlot{};
+        head                = (head + 1) % kDeferredQueueCapacity;
+        --count;
+        return out;
+    }
+};
+
+/** Swap ring state (head, count) and per-slot storage without copying whole FixedBuffer (avoids copy-only path). */
+void SwapDeferredRings(DeferredUdpRing & a, DeferredUdpRing & b)
+{
+    using std::swap;
+    swap(a.head, b.head);
+    swap(a.count, b.count);
+    for (size_t i = 0; i < kDeferredQueueCapacity; ++i)
+    {
+        swap(a.slots[i], b.slots[i]);
+    }
+}
+
+DeferredUdpRing gDeferredRing;
+
+bool IsNetifUsable(struct netif * netif)
+{
+    return netif != nullptr && netif_is_up(netif) && netif_is_link_up(netif);
+}
+
+#if LWIP_IPV6
+bool NetifHasValidIpv6Address(struct netif * netif)
+{
+    VerifyOrReturnValue(netif != nullptr, false);
+    for (u8_t idx = 0; idx < LWIP_IPV6_NUM_ADDRESSES; ++idx)
+    {
+        if (ip6_addr_ispreferred(netif_ip6_addr_state(netif, idx)))
+        {
+            return true;
+        }
+    }
+    return false;
+}
+#endif
+
+bool IsNetifReadyForOutboundUdp(struct netif * netif, const IPAddress & dest)
+{
+    VerifyOrReturnValue(IsNetifUsable(netif), false);
+#if LWIP_IPV6
+    if (dest.IsIPv6())
+    {
+        return NetifHasValidIpv6Address(netif);
+    }
+#else
+    if (dest.IsIPv6())
+    {
+        return false;
+    }
+#endif
+#if INET_CONFIG_ENABLE_IPV4 && LWIP_IPV4
+    if (dest.IsIPv4())
+    {
+        return !ip4_addr_isany(netif_ip4_addr(netif));
+    }
+#endif
+    return false;
+}
+
+bool IsOutboundNetifReadyForUdp(const IPPacketInfo & pktInfo)
+{
+    const InterfaceId & intfId  = pktInfo.Interface;
+    const IPAddress & dest     = pktInfo.DestAddress;
+
+    if (intfId.IsPresent())
+    {
+        return IsNetifReadyForOutboundUdp(intfId.GetPlatformInterface(), dest);
+    }
+
+    if (IsNetifReadyForOutboundUdp(netif_default, dest))
+    {
+        return true;
+    }
+
+    struct netif * netif = nullptr;
+    NETIF_FOREACH(netif)
+    {
+        if (IsNetifReadyForOutboundUdp(netif, dest))
+        {
+            return true;
+        }
+    }
+
+    return false;
+}
+
+} // namespace
+
+CHIP_ERROR DeferredUdpSendQueueLwIP::ProbeDefer(const IPPacketInfo & pktInfo, bool & outShouldDefer)
+{
+    outShouldDefer  = false;
+    err_t probeErr = EndPointStateLwIP::RunOnTCPIPRet([&]() -> err_t {
+        outShouldDefer = !IsOutboundNetifReadyForUdp(pktInfo);
+        return ERR_OK;
+    });
+    VerifyOrReturnError(probeErr == ERR_OK, chip::System::MapErrorLwIP(probeErr));
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR DeferredUdpSendQueueLwIP::Enqueue(UDPEndPointImplLwIP * self, const IPPacketInfo * pktInfo, System::PacketBufferHandle && msg)
+{
+    DeferredUdpSlot slot;
+    slot.epHandle = UDPEndPointHandle(static_cast<UDPEndPoint *>(self));
+    slot.pktInfo  = *pktInfo;
+    slot.msg      = std::move(msg);
+
+    char destStr[IPAddress::kMaxStringLength];
+    pktInfo->DestAddress.ToString(destStr);
+    ChipLogProgress(Inet, "UDP send deferred (waiting for netif): dest %s port %u len %u", destStr, pktInfo->DestPort,
+                    static_cast<unsigned>(slot.msg->TotalLength()));
+
+    gDeferredRing.PushBack(std::move(slot));
+    return CHIP_NO_ERROR;
+}
+
+void DeferredUdpSendQueueLwIP::PurgeForEndpoint(UDPEndPointImplLwIP * self)
+{
+    UDPEndPoint * asBase = static_cast<UDPEndPoint *>(self);
+    const size_t n       = gDeferredRing.count;
+    for (size_t i = 0; i < n; ++i)
+    {
+        DeferredUdpSlot slot = gDeferredRing.PopFront();
+        if (!slot.epHandle.IsNull() && slot.epHandle == *asBase)
+        {
+            continue;
+        }
+        gDeferredRing.PushBack(std::move(slot));
+    }
+}
+
+void DeferredUdpSendQueueLwIP::Flush()
+{
+    assertChipStackLockedByCurrentThread();
+
+    DeferredUdpRing pending{};
+    SwapDeferredRings(gDeferredRing, pending);
+
+    while (!pending.Empty())
+    {
+        DeferredUdpSlot slot = pending.PopFront();
+
+        bool ready     = false;
+        err_t probeErr = EndPointStateLwIP::RunOnTCPIPRet([&]() -> err_t {
+            ready = IsOutboundNetifReadyForUdp(slot.pktInfo);
+            return ERR_OK;
+        });
+        if (probeErr != ERR_OK)
+        {
+            ChipLogError(Inet, "DeferredUdpSendQueueLwIP::Flush: RunOnTCPIPRet failed: %d", static_cast<int>(probeErr));
+            gDeferredRing.PushBack(std::move(slot));
+            while (!pending.Empty())
+            {
+                gDeferredRing.PushBack(pending.PopFront());
+            }
+            return;
+        }
+
+        if (!ready)
+        {
+            gDeferredRing.PushBack(std::move(slot));
+            continue;
+        }
+
+        if (!slot.epHandle.IsNull())
+        {
+            auto * impl = static_cast<UDPEndPointImplLwIP *>(slot.epHandle.operator->());
+            (void) impl->FlushOneDeferred(&slot.pktInfo, std::move(slot.msg));
+        }
+    }
+}
+
+} // namespace Inet
+} // namespace chip
+
+#endif // INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY
--- a/src/inet/DeferredUdpSendQueueLwIP.cpp
+++ b/src/inet/DeferredUdpSendQueueLwIP.cpp
@@ -68,11 +68,7 @@ struct DeferredUdpRing
     {
         if (count == kDeferredQueueCapacity)
         {
-            const DeferredUdpSlot & dropped = slots[head];
-            char dropDest[IPAddress::kMaxStringLength];
-            dropped.pktInfo.DestAddress.ToString(dropDest);
-            ChipLogProgress(Inet, "Deferred UDP queue full: dropping head dest %s port %u len %u", dropDest,
-                            dropped.pktInfo.DestPort, static_cast<unsigned>(dropped.msg->TotalLength()));
+            ChipLogDetail(Inet, "Deferred UDP queue full: dropping head");
             slots[head] = DeferredUdpSlot{};
             head        = (head + 1) % kDeferredQueueCapacity;
             --count;
@@ -111,7 +107,7 @@ bool IsNetifUsable(struct netif * netif)
     return netif != nullptr && netif_is_up(netif) && netif_is_link_up(netif);
 }
 
-bool NetifHasValidIpv6Address(struct netif * netif)
+bool NetifHasPreferredIpv6Address(struct netif * netif)
 {
     VerifyOrReturnValue(netif != nullptr, false);
     for (u8_t idx = 0; idx < LWIP_IPV6_NUM_ADDRESSES; ++idx)
@@ -129,7 +125,7 @@ bool IsNetifReadyForOutboundUdp(struct netif * netif, const IPAddress & dest)
     VerifyOrReturnValue(IsNetifUsable(netif), false);
     if (dest.IsIPv6())
     {
-        return NetifHasValidIpv6Address(netif);
+        return NetifHasPreferredIpv6Address(netif);
     }
 #if INET_CONFIG_ENABLE_IPV4 && LWIP_IPV4
     if (dest.IsIPv4())
@@ -188,10 +184,7 @@ CHIP_ERROR DeferredUdpSendQueueLwIP::Enqueue(UDPEndPointImplLwIP * self, const I
     slot.pktInfo  = *pktInfo;
     slot.msg      = std::move(msg);
 
-    char destStr[IPAddress::kMaxStringLength];
-    pktInfo->DestAddress.ToString(destStr);
-    ChipLogProgress(Inet, "UDP send deferred (waiting for netif): dest %s port %u len %u", destStr, pktInfo->DestPort,
-                    static_cast<unsigned>(slot.msg->TotalLength()));
+    ChipLogDetail(Inet, "UDP send deferred (waiting for netif)");
 
     gDeferredRing.PushBack(std::move(slot));
     return CHIP_NO_ERROR;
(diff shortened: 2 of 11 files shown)


src/inet/InetConfig.h
rosahay-silabs, line 1: In order to reduce the conflicts, can we move this to `src/inet/InetConfigExtended.h` and import that in our custom `UDPEndPointImplLwIP.cpp` only.

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -57,6 +63,148 @@ static_assert(LWIP_VERSION_MAJOR > 1, "CHIP requires LwIP 2.0 or later");
...
+#if LWIP_IPV6
+bool NetifHasValidIpv6Address(struct netif * netif)
+{
+    if (netif == nullptr)
+    {
+        return false;
+    }
+    for (u8_t idx = 0; idx < LWIP_IPV6_NUM_ADDRESSES; ++idx)
+    {
+        if (ip6_addr_isvalid(netif_ip6_addr_state(netif, idx)))
```
rosahay-silabs: This condition can still give a LwIP routing error as per my understanding, should we be using `ip6_addr_ispreferred` instead?

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -57,6 +63,148 @@ static_assert(LWIP_VERSION_MAJOR > 1, "CHIP requires LwIP 2.0 or later");
...
+}
+#endif // LWIP_IPV6
+
+bool IsNetifReadyForOutboundUdp(struct netif * netif, const IPAddress & dest)
+{
+    if (!IsNetifUsable(netif))
+    {
+        return false;
+    }
+#if LWIP_IPV6
```
rosahay-silabs: considering `LWIP_IPV6` is not optional, is this required?

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -57,6 +63,148 @@ static_assert(LWIP_VERSION_MAJOR > 1, "CHIP requires LwIP 2.0 or later");
...
+// FIFO of deferred UDP sends, bounded to INET_CONFIG_UDP_LWIP_DEFERRED_SEND_QUEUE_SIZE.
+// Enqueue drops the oldest entry on overflow; purge/flush erase matching entries in place.
+std::deque<DeferredUdpSlot> gDeferredUdpQueue;
+
+bool IsNetifUsable(struct netif * netif)
+{
+    return netif != nullptr && netif_is_up(netif) && netif_is_link_up(netif);
+}
+
+#if LWIP_IPV6
```
rosahay-silabs: considering `LWIP_IPV6` is not optional, is this required?

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -57,6 +63,148 @@ static_assert(LWIP_VERSION_MAJOR > 1, "CHIP requires LwIP 2.0 or later");
...
+    {
+        return IsNetifReadyForOutboundUdp(intfId.GetPlatformInterface(), dest);
+    }
+
+    if (IsNetifReadyForOutboundUdp(netif_default, dest))
+    {
+        return true;
+    }
+
+#if defined(NETIF_FOREACH)
```
rosahay-silabs: the following logic can be simplified since the loop body is the same.

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -57,6 +63,148 @@ static_assert(LWIP_VERSION_MAJOR > 1, "CHIP requires LwIP 2.0 or later");
 namespace chip {
 namespace Inet {
 
+#if INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY
+
+namespace {
+
+struct DeferredUdpSlot
+{
+    UDPEndPointHandle ep;
```
rosahay-silabs:
```suggestion
    UDPEndPointHandle epHandle;
```

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -57,6 +63,148 @@ static_assert(LWIP_VERSION_MAJOR > 1, "CHIP requires LwIP 2.0 or later");
...
+}
+
+CHIP_ERROR EnqueueDeferredUdpSend(UDPEndPointImplLwIP * self, const IPPacketInfo * pktInfo, System::PacketBufferHandle && msg)
+{
+    if (gDeferredUdpQueue.size() >= INET_CONFIG_UDP_LWIP_DEFERRED_SEND_QUEUE_SIZE)
+    {
+        const DeferredUdpSlot & head = gDeferredUdpQueue.front();
+        char dropDest[IPAddress::kMaxStringLength];
+        head.pktInfo.DestAddress.ToString(dropDest);
+        ChipLogProgress(Inet, "Deferred UDP queue full: dropping head dest %s port %u len %u", dropDest,
```
rosahay-silabs:
```suggestion
        ChipLogDetail(Inet, "Deferred UDP queue full: dropping head dest %s port %u len %u", dropDest,
```

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -57,6 +63,148 @@ static_assert(LWIP_VERSION_MAJOR > 1, "CHIP requires LwIP 2.0 or later");
...
+    }
+
+    DeferredUdpSlot slot;
+    slot.ep      = UDPEndPointHandle(static_cast<UDPEndPoint *>(self));
+    slot.pktInfo = *pktInfo;
+    slot.msg     = std::move(msg);
+
+    char destStr[IPAddress::kMaxStringLength];
+    pktInfo->DestAddress.ToString(destStr);
+    ChipLogProgress(Inet, "UDP send deferred (waiting for netif): dest %s port %u len %u", destStr, pktInfo->DestPort,
```
rosahay-silabs:
```suggestion
    ChipLogDetail(Inet, "UDP send deferred (waiting for netif): dest %s port %u len %u", destStr, pktInfo->DestPort,
```

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -57,6 +63,148 @@ static_assert(LWIP_VERSION_MAJOR > 1, "CHIP requires LwIP 2.0 or later");
...
+                    static_cast<unsigned>(slot.msg->TotalLength()));
+
+    gDeferredUdpQueue.push_back(std::move(slot));
```
rosahay-silabs: what does `std::move(slot)` signify while doing `push_back`?
like how is this better than `gDeferredUdpQueue.push_back(slot);`

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -57,6 +63,148 @@ static_assert(LWIP_VERSION_MAJOR > 1, "CHIP requires LwIP 2.0 or later");
...
+        head.pktInfo.DestAddress.ToString(dropDest);
+        ChipLogProgress(Inet, "Deferred UDP queue full: dropping head dest %s port %u len %u", dropDest,
+                        head.pktInfo.DestPort, static_cast<unsigned>(head.msg->TotalLength()));
+        gDeferredUdpQueue.pop_front();
+    }
+
+    DeferredUdpSlot slot;
+    slot.ep      = UDPEndPointHandle(static_cast<UDPEndPoint *>(self));
+    slot.pktInfo = *pktInfo;
+    slot.msg     = std::move(msg);
```
rosahay-silabs: `std::move(msg)` is this actually saving anything compared to just copying the `msg`?

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -57,6 +63,148 @@ static_assert(LWIP_VERSION_MAJOR > 1, "CHIP requires LwIP 2.0 or later");
...
+    ChipLogProgress(Inet, "UDP send deferred (waiting for netif): dest %s port %u len %u", destStr, pktInfo->DestPort,
+                    static_cast<unsigned>(slot.msg->TotalLength()));
+
+    gDeferredUdpQueue.push_back(std::move(slot));
+    return CHIP_NO_ERROR;
+}
+
+void PurgeDeferredUdpSendsForEndpoint(UDPEndPointImplLwIP * self)
+{
+    UDPEndPoint * asBase = static_cast<UDPEndPoint *>(self);
```
rosahay-silabs: `gDeferredUdpQueue.clear()` should purge the entire container, any reason to verify `!it->ep.IsNull() && it->ep == *asBase` ?

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -192,18 +363,91 @@ CHIP_ERROR UDPEndPointImplLwIP::SendMsgImpl(const IPPacketInfo * pktInfo, System
...
     ip_addr_copy(mUDP->local_ip, boundAddr);
 
+    const unsigned payloadLen = static_cast<unsigned>(msg->TotalLength());
+    char destStr[IPAddress::kMaxStringLength];
+    destAddr.ToString(destStr);
+
+#if INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY
+    if (lwipErr == ERR_RTE)
+    {
+        ChipLogProgress(Inet, "UDP send deferred (no route yet, ERR_RTE): dest %s port %u len %u", destStr, destPort, payloadLen);
```
rosahay-silabs:
```suggestion
        ChipLogDetail(Inet, "UDP send deferred (no route yet, ERR_RTE): dest %s port %u len %u", destStr, destPort, payloadLen);
```

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -192,18 +363,91 @@ CHIP_ERROR UDPEndPointImplLwIP::SendMsgImpl(const IPPacketInfo * pktInfo, System
...
+#if INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY
+void UDPEndPointImplLwIP::FlushDeferredSendQueue()
+{
+    assertChipStackLockedByCurrentThread();
+
+    // Take ownership of the current queue locally. This guarantees each entry is visited
+    // exactly once per flush call, and is safe if PerformLwIPUdpSend re-enqueues on ERR_RTE
+    // (those go into the now-empty gDeferredUdpQueue without affecting our iteration).
+    std::deque<DeferredUdpSlot> pending;
+    pending.swap(gDeferredUdpQueue);
```
rosahay-silabs: what is the use of `swap` here? are we clearing out the `gDeferredUdpQueue` in this stage?

src/inet/InetConfig.h
chirag-silabs: Yes, I have kept these here for time being for the diff between UDPEndpointLwIP.cpp and the new file i.e., UDPEndpointsPlatform.cpp can be seen easily
These will be moved once to a different file once the changes are all good.

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: Keeping it in a local list to ensure adding it back if send doesn't happens

image

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: Yes will be moving this, keeping this for the testing.

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: The same was used in the LwIP files so used the same.

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: std::move is used every where in the UDPEndpointLwIP.cpp file since DeferredUdpSlot Struct consist of `System::PacketBufferHandle` which is explicity non-copyable i.e., assignment are deleted

`private:`
`   PacketBufferHandle(const PacketBufferHandle &)             = delete;`
`    PacketBufferHandle & operator=(const PacketBufferHandle &) = delete;`

src/inet/UDPEndPointImplLwIP.cpp
rosahay-silabs: Resolve the log suggestions together when you revert the logs.

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -643,11 +635,11 @@ void UDPEndPointImplLwIP::LwIPReceiveUDPMessage(void * arg, struct udp_pcb * pcb
         // the packet is basically dequeued, if it tries to keep track of the lifecycle.
         if (sQueueFilter != nullptr)
         {
-            (void) sQueueFilter->FilterAfterDequeue(ep, *(pktInfo.get()), buf);
+            (void) sQueueFilter->FilterAfterDequeue(epHandle, *(pktInfo.get()), buf);
```
rosahay-silabs:
```suggestion
            RETURN_SAFELY_IGNORED sQueueFilter->FilterAfterDequeue(epHandle, *(pktInfo.get()), buf);
```

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -426,9 +418,9 @@ void UDPEndPointImplLwIP::FlushDeferredSendQueue()
...
         }
 
-        if (!slot.ep.IsNull())
+        if (!slot.epHandle.IsNull())
         {
-            auto * impl = static_cast<UDPEndPointImplLwIP *>(slot.ep.operator->());
+            auto * impl = static_cast<UDPEndPointImplLwIP *>(slot.epHandle.operator->());
             if (impl->mState != State::kClosed && impl->mUDP != nullptr)
             {
                 (void) impl->PerformLwIPUdpSend(&slot.pktInfo, std::move(slot.msg));
```
rosahay-silabs:
```suggestion
                RETURN_SAFELY_IGNORED impl->PerformLwIPUdpSend(&slot.pktInfo, std::move(slot.msg));
```

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: Keeping this for time being for debugging

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: Keeping this for time being for debugging

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -57,6 +63,140 @@ static_assert(LWIP_VERSION_MAJOR > 1, "CHIP requires LwIP 2.0 or later");
 namespace chip {
 namespace Inet {
 
+#if INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY
+
+namespace {
+
+// Fallback for LwIP versions that do not provide NETIF_FOREACH (pre-2.1 or with LWIP_SINGLE_NETIF).
+// Matches the upstream signature: the caller declares the loop variable.
+#ifndef NETIF_FOREACH
```
rosahay-silabs: fine for now, but it should be moved into a common part if this is not present, anywhere else?

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -57,6 +63,140 @@ static_assert(LWIP_VERSION_MAJOR > 1, "CHIP requires LwIP 2.0 or later");
...
+// FIFO of deferred UDP sends, bounded to INET_CONFIG_UDP_LWIP_DEFERRED_SEND_QUEUE_SIZE.
+// Enqueue drops the oldest entry on overflow; purge/flush erase matching entries in place.
+std::deque<DeferredUdpSlot> gDeferredUdpQueue;
+
+bool IsNetifUsable(struct netif * netif)
+{
+    return netif != nullptr && netif_is_up(netif) && netif_is_link_up(netif);
+}
+
+bool NetifHasValidIpv6Address(struct netif * netif)
```
rosahay-silabs:
```suggestion
bool NetifHasPreferredIpv6Address(struct netif * netif)
```

rosahay-silabs (approved):
After addressing the comments rest look good to me

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -23,10 +23,16 @@
 
 #include <inet/UDPEndPointImplLwIP.h>
 
+#include <deque>
```
jmartinez-silabs: I didn't know about this lib. It might be ok to use but I have some worries on code size impact and run time impact.
This lib use polymorphic_allocator and I assume multiple dynamic memory allocation during queuing/dequeuing operation. This is typically not well suited form small, memory-constrained, embedded MCU.
https://en.cppreference.com/cpp/container/deque

Matter has some existing queue/list class implementations that are light weight. It might be valuable to look at those alternatives.

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -150,38 +290,61 @@ CHIP_ERROR UDPEndPointImplLwIP::SendMsgImpl(const IPPacketInfo * pktInfo, System
...
     // Send the message to the specified address/port.
     // If an outbound interface has been specified, call a specific version of the UDP sendto()
     // function that accepts the target interface.
     // If a source address has been specified, temporarily override the local_ip of the PCB.
     // This results in LwIP using the given address being as the source address for the generated
     // packet, as if the PCB had been bound to that address.
-    const IPAddress & srcAddr  = pktInfo->SrcAddress;
-    const uint16_t & destPort  = pktInfo->DestPort;
+    const IPAddress & srcAddr   = pktInfo->SrcAddress;
+    const uint16_t & destPort    = pktInfo->DestPort;
```
jmartinez-silabs: nit: restyle should have kept this align like it was before

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -325,8 +561,8 @@ CHIP_ERROR UDPEndPointImplLwIP::GetPCB(IPAddressType addrType)
 void UDPEndPointImplLwIP::LwIPReceiveUDPMessage(void * arg, struct udp_pcb * pcb, struct pbuf * p, const ip_addr_t * addr,
                                                 u16_t port)
 {
-    UDPEndPointImplLwIP * ep = static_cast<UDPEndPointImplLwIP *>(arg);
-    if (ep->mState == State::kClosed)
+    UDPEndPointImplLwIP * epHandle = static_cast<UDPEndPointImplLwIP *>(arg);
```
jmartinez-silabs: Since this is a file common in CSA. careful with the change that are just for the sake of change if you aren't going to upstream it.

Limiting the change to what is needed will reduce the change of conflict and possible issues with badly applied resolutions

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -57,6 +63,140 @@ static_assert(LWIP_VERSION_MAJOR > 1, "CHIP requires LwIP 2.0 or later");
...
+        {
+            ++it;
+        }
+    }
+}
+
+} // namespace
+
+#endif // INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY
+
```
jmartinez-silabs: Could this added code for the new queuw mechanism be done in a new file and class of its own and used as utility class for our use case?

This would separate our code additions from the common csa file nicely

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -150,38 +290,61 @@ CHIP_ERROR UDPEndPointImplLwIP::SendMsgImpl(const IPPacketInfo * pktInfo, System
...
-    // Make sure we have the appropriate type of PCB based on the destination address.
-    res = GetPCB(destAddr.Type());
+    CHIP_ERROR res = GetPCB(destAddr.Type());
     if (res != CHIP_NO_ERROR)
     {
         return res;
     }
 
+#if INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY
+    FlushDeferredSendQueue();
```
jmartinez-silabs: why do we flush the queue here?

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -150,38 +290,61 @@ CHIP_ERROR UDPEndPointImplLwIP::SendMsgImpl(const IPPacketInfo * pktInfo, System
...
+
+    bool defer = false;
+    err_t deferProbeErr = RunOnTCPIPRet([&]() -> err_t {
+        defer = !IsOutboundNetifReadyForUdp(*pktInfo);
+        return ERR_OK;
+    });
+    if (deferProbeErr != ERR_OK)
+    {
+        return chip::System::MapErrorLwIP(deferProbeErr);
+    }
```
jmartinez-silabs:
```suggestion
    VerifyOrReturnError(deferProbeErr == ERR_OK, chip::System::MapErrorLwIP(deferProbeErr));
```

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -192,18 +355,91 @@ CHIP_ERROR UDPEndPointImplLwIP::SendMsgImpl(const IPPacketInfo * pktInfo, System
...
+    while (!pending.empty())
+    {
+        DeferredUdpSlot slot = std::move(pending.front());
+        pending.pop_front();
+
+        bool ready         = false;
+        err_t probeErr = RunOnTCPIPRet([&]() -> err_t {
+            ready = IsOutboundNetifReadyForUdp(slot.pktInfo);
+            return ERR_OK;
+        });
```
jmartinez-silabs: Should we not check this (IsOutboundNetifReadyForUdp) before moving to the local
```
std::deque<DeferredUdpSlot> pending;
pending.swap(gDeferredUdpQueue);
```
and at least before the 
```
DeferredUdpSlot slot = std::move(pending.front());
pending.pop_front();
```

IMO it should be check if the network interface is ready then manipulate the queue.
No need to play with the queue if we won't be able to send no?

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -150,38 +290,61 @@ CHIP_ERROR UDPEndPointImplLwIP::SendMsgImpl(const IPPacketInfo * pktInfo, System
...
     }
 
+#if INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY
+    FlushDeferredSendQueue();
+
+    bool defer = false;
+    err_t deferProbeErr = RunOnTCPIPRet([&]() -> err_t {
+        defer = !IsOutboundNetifReadyForUdp(*pktInfo);
+        return ERR_OK;
+    });
```
jmartinez-silabs: In `FlushDeferredSendQueue` we check  `IsOutboundNetifReadyForUdp` we get out of FlushDeferredSendQueue and check it again. That seem redundant and that we are calling functions maybe for no reasons.

My understanding of what should happen here (in simplified pseudo code) :
```
if Netif is ready == true:
     FlushDeferredSendQueue();
     PerformLwIPUdpSend();
else :
    EnqueueDeferredUdpSend();
```
     
and FlushDeferredSendQueue shouldn't have to check if netIf is ready again.

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -57,6 +63,140 @@ static_assert(LWIP_VERSION_MAJOR > 1, "CHIP requires LwIP 2.0 or later");
...
+        {
+            return true;
+        }
+    }
+    return false;
+}
+
+bool IsNetifReadyForOutboundUdp(struct netif * netif, const IPAddress & dest)
+{
+    if (!IsNetifUsable(netif))
```
jmartinez-silabs: `VerifyOrReturnValue` could be use here for nice 1 liners for all those checks.

jmartinez-silabs: `VerifyOrReturnValue` could be use here for nice 1 liners for all those checks.

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -57,6 +63,140 @@ static_assert(LWIP_VERSION_MAJOR > 1, "CHIP requires LwIP 2.0 or later");
...
+    {
+        return NetifHasValidIpv6Address(netif);
+    }
+#if INET_CONFIG_ENABLE_IPV4 && LWIP_IPV4
+    if (dest.IsIPv4())
+    {
+        return !ip4_addr_isany(netif_ip4_addr(netif));
+    }
+#endif // INET_CONFIG_ENABLE_IPV4 && LWIP_IPV4
+    return true;
```
jmartinez-silabs: should this default to true?
If (dest.IsIPv6()) is false and we don't support Ipv4? Should we say that the interface is ready?

chirag-silabs: changed the by default return to false.

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: Correct, it was duplicated. Updated it

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: Updated, before flushing the check is already added to handle these cases.

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: Added a new file for the queue handling.

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: Removed the no used changes.

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: updated.

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: Updated to use the supported src/lib/support/FixedBuffer.h for this handling since we don't have the support for queue or deque in CSA support files.

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: Updated

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: Done.

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: selective purge will be correct; full clear on one CloseImpl would break other endpoints’ queued traffic. Keep this generic

src/inet/DeferredUdpSendQueueLwIP.h
```diff
@@ -0,0 +1,57 @@
+/*
+ *
+ *    Copyright (c) 2020-2025 Project CHIP Authors
```
rosahay-silabs:
```suggestion
 *    Copyright (c) 2020-2026 Project CHIP Authors
```

src/inet/DeferredUdpSendQueueLwIP.cpp
```diff
@@ -0,0 +1,252 @@
+/*
+ *
+ *    Copyright (c) 2020-2025 Project CHIP Authors
```
rosahay-silabs:
```suggestion
 *    Copyright (c) 2020-2026 Project CHIP Authors
```

src/inet/DeferredUdpSendQueueLwIP.cpp
```diff
@@ -0,0 +1,252 @@
...
+    if (dest.IsIPv6())
+    {
+        return NetifHasPreferredIpv6Address(netif);
+    }
+#if INET_CONFIG_ENABLE_IPV4 && LWIP_IPV4
+    if (dest.IsIPv4())
+    {
+        return !ip4_addr_isany(netif_ip4_addr(netif));
+    }
+#endif
```
rosahay-silabs:
```suggestion
#endif // INET_CONFIG_ENABLE_IPV4 && LWIP_IPV4
```

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -150,16 +154,33 @@ CHIP_ERROR UDPEndPointImplLwIP::SendMsgImpl(const IPPacketInfo * pktInfo, System
         VerifyOrReturnError(!msg.IsNull(), CHIP_ERROR_NO_MEMORY);
     }
 
-    CHIP_ERROR res = CHIP_NO_ERROR;
-    err_t lwipErr  = ERR_VAL;
-
-    // Make sure we have the appropriate type of PCB based on the destination address.
-    res = GetPCB(destAddr.Type());
+    CHIP_ERROR res = GetPCB(destAddr.Type());
```
rosahay-silabs: Either upstream this change or revert this change for now to reduce the conflicts going forward

src/inet/UDPEndPointImplLwIP.cpp
rosahay-silabs, line 1: Why did the permission of this file change? though I agree it should be `644`

src/inet/DeferredUdpSendQueueLwIP.cpp
```diff
@@ -0,0 +1,252 @@
...
+}
+
+bool IsNetifReadyForOutboundUdp(struct netif * netif, const IPAddress & dest)
+{
+    VerifyOrReturnValue(IsNetifUsable(netif), false);
+    if (dest.IsIPv6())
+    {
+        return NetifHasPreferredIpv6Address(netif);
+    }
+#if INET_CONFIG_ENABLE_IPV4 && LWIP_IPV4
```
rosahay-silabs: `INET_CONFIG_ENABLE_IPV4` should be enough going by the use in other APIs

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -181,7 +202,7 @@ CHIP_ERROR UDPEndPointImplLwIP::SendMsgImpl(const IPPacketInfo * pktInfo, System
         ip_addr_copy(mUDP->local_ip, lwipSrcAddr);
     }
 
-    lwipErr = RunOnTCPIPRet([this, &intfId, &msg, &lwipDestAddr, destPort]() {
+    err_t lwipErr = RunOnTCPIPRet([this, &intfId, &msg, &lwipDestAddr, destPort]() {
```
rosahay-silabs: instead of changing it here, again in order to reduce the conflicts move `err_t lwipErr` to the top of the function body where we are doing the changes

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -150,16 +154,33 @@ CHIP_ERROR UDPEndPointImplLwIP::SendMsgImpl(const IPPacketInfo * pktInfo, System
         VerifyOrReturnError(!msg.IsNull(), CHIP_ERROR_NO_MEMORY);
     }
 
-    CHIP_ERROR res = CHIP_NO_ERROR;
-    err_t lwipErr  = ERR_VAL;
```
rosahay-silabs, line 154: Just removal of this line should be better than changing both the lines

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -150,16 +154,33 @@ CHIP_ERROR UDPEndPointImplLwIP::SendMsgImpl(const IPPacketInfo * pktInfo, System
...
+#endif // INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY
+
+    return PerformLwIPUdpSend(pktInfo, std::move(msg));
+}
+
+CHIP_ERROR UDPEndPointImplLwIP::PerformLwIPUdpSend(const IPPacketInfo * pktInfo, System::PacketBufferHandle && msg)
+{
+    assertChipStackLockedByCurrentThread();
+
+    const IPAddress & destAddr = pktInfo->DestAddress;
```
rosahay-silabs, line 188: add somewhere here `err_t lwipErr = ERR_VAL;`

src/inet/UDPEndPointImplLwIP.cpp
```diff
@@ -192,18 +213,47 @@ CHIP_ERROR UDPEndPointImplLwIP::SendMsgImpl(const IPPacketInfo * pktInfo, System
...
+    {
+        ChipLogDetail(Inet, "UDP send deferred (no route yet, ERR_RTE)");
+        return DeferredUdpSendQueueLwIP::Enqueue(this, pktInfo, std::move(msg));
+    }
+#endif // INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY
+
     if (lwipErr != ERR_OK)
     {
-        res = chip::System::MapErrorLwIP(lwipErr);
+        return chip::System::MapErrorLwIP(lwipErr);
```
rosahay-silabs, line 232: can we consolidate the changes?

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: added back

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: added

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: Common code, hence not modifying this.

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: same as above, it is common code hence not modifying

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: moved to our file.

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: Updated.

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: reverted.

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: No idea :D

src/inet/UDPEndPointImplLwIP.cpp
chirag-silabs: not this one since res is not a scope variable of this function.

src/inet/BUILD.gn
```diff
@@ -50,6 +50,10 @@ buildconfig_header("inet_buildconfig") {
     "INET_CONFIG_ENABLE_TCP_ENDPOINT=${chip_inet_config_enable_tcp_endpoint}",
     "INET_CONFIG_ENABLE_UDP_ENDPOINT=${chip_inet_config_enable_udp_endpoint}",
     "HAVE_LWIP_RAW_BIND_NETIF=true",
+
+    # [SL-ONLY] Queuing the UDP packets for till the netif is not up
```
jmartinez-silabs:
```suggestion
    # [SL-ONLY] Queuing the UDP packets for until the netif is up
```

src/inet/DeferredUdpSendQueueLwIP.cpp
```diff
@@ -0,0 +1,252 @@
...
+        head                = (head + 1) % kDeferredQueueCapacity;
+        --count;
+        return out;
+    }
+};
+
+/** Swap ring state (head, count) and per-slot storage without copying whole FixedBuffer (avoids copy-only path). */
+void SwapDeferredRings(DeferredUdpRing & a, DeferredUdpRing & b)
+{
+    using std::swap;
```
jmartinez-silabs: This using is a bit overkill
It only saves std::

EtienneCampeauSilabs (commented):
This is a new feature. Your new class could be unit tested

src/inet/DeferredUdpSendQueueLwIP.cpp
```diff
@@ -0,0 +1,252 @@
...
+namespace chip {
+namespace Inet {
+
+namespace {
+
+#ifndef NETIF_FOREACH
+#define NETIF_FOREACH(netif) for ((netif) = netif_list; (netif) != nullptr; (netif) = (netif)->next)
+#endif
+
+constexpr size_t kDeferredQueueCapacity = SILABS_INET_CONFIG_UDP_LWIP_DEFERRED_SEND_QUEUE_SIZE;
```
jmartinez-silabs: nit, we typically use SL_ for our defines
```suggestion
constexpr size_t kDeferredQueueCapacity = SL_INET_CONFIG_UDP_LWIP_DEFERRED_SEND_QUEUE_SIZE;
```

src/inet/DeferredUdpSendQueueLwIP.cpp
```diff
@@ -0,0 +1,252 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+#include <inet/DeferredUdpSendQueueLwIP.h>
+
+#if SILABS_INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY
```
jmartinez-silabs:
```suggestion
#if SL_INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY
```

src/inet/DeferredUdpSendQueueLwIP.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+    DeferredUdpSlot PopFront()
+    {
+        DeferredUdpSlot out = std::move(slots[head]);
+        slots[head]         = DeferredUdpSlot{};
+        head                = (head + 1) % kDeferredQueueCapacity;
+        --count;
+        return out;
+    }
+};
```
jmartinez-silabs: You seem to be mixing up both queue and circular/ring buffer in one concept. 

One or the other would work but here, you are adding extra steps that wouldn't be needed. For example, in a ringbuffer there is no PopFront/PushBack but just a head and tail index to write or read and you don't have to "clear the content of the index being read. 

I am also not sure that this `DeferredUdpSlot out`... `return out;` works. isn't that a local struct on the stack which memory can be re-used once out of the PopFront function?

Not 100% sure maybe it works but it would highly benefit from being unit tests.

src/inet/DeferredUdpSendQueueLwIP.cpp
```diff
@@ -0,0 +1,252 @@
...
+    using std::swap;
+    swap(a.head, b.head);
+    swap(a.count, b.count);
+    for (size_t i = 0; i < kDeferredQueueCapacity; ++i)
+    {
+        swap(a.slots[i], b.slots[i]);
+    }
+}
+
+DeferredUdpRing gDeferredRing;
```
jmartinez-silabs, line 61: this `gDeferredRing` is global  and i see that DeferredUdpSlot can have different UDP `UDPEndPointHandle epHandle;`. Is this thread-safe? Will multiple thread try to use gDeferredRing?

src/inet/EndPointStateLwIP.h
```diff
@@ -33,11 +33,18 @@ struct tcp_pcb;
...
+class DeferredUdpSendQueueLwIP;
+#endif
+
 /**
  * Definitions shared by all LwIP EndPoint classes.
  */
 class DLL_EXPORT EndPointStateLwIP
 {
+#if SILABS_INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY
+    friend class DeferredUdpSendQueueLwIP;
```
jmartinez-silabs, line 46: why does it need to be friend?

src/inet/inet.gni
```diff
@@ -24,6 +24,12 @@ declare_args() {
   # Enable UDP endpoint.
   chip_inet_config_enable_udp_endpoint = true
 
+  # [SL-ONLY]: defer outbound UDP until a netif is link-up (DeferredUdpSendQueueLwIP).
+  silabs_inet_config_udp_lwip_queue_until_netif_ready = false
+
+  # [SL-ONLY]: queue size for the UDP DeferredUdpSendQueueLwIP
+  silabs_inet_config_udp_lwip_deferred_send_queue_size = 0
```
jmartinez-silabs:
```suggestion
  # [SL-ONLY]: defer outbound UDP until a netif is link-up (DeferredUdpSendQueueLwIP).
  sl_inet_config_udp_lwip_queue_until_netif_ready = false

  # [SL-ONLY]: queue size for the UDP DeferredUdpSendQueueLwIP
  sl_inet_config_udp_lwip_deferred_send_queue_size = 0
```

src/inet/UDPEndPointImplLwIP.h
```diff
@@ -30,8 +30,16 @@
...
 namespace Inet {
 
+#if SILABS_INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY
+class DeferredUdpSendQueueLwIP;
+#endif
+
 class UDPEndPointImplLwIP : public UDPEndPoint, public EndPointStateLwIP
 {
+#if SILABS_INET_CONFIG_UDP_LWIP_QUEUE_UNTIL_NETIF_READY
+    friend class DeferredUdpSendQueueLwIP;
```
jmartinez-silabs, line 40: similar question why does it need to be a friend class?

src/inet/UDPEndPointImplLwIP.h
```diff
@@ -54,6 +62,12 @@ class UDPEndPointImplLwIP : public UDPEndPoint, public EndPointStateLwIP
      */
     static void SetQueueFilter(EndpointQueueFilter * queueFilter) { sQueueFilter = queueFilter; }
 
+    /**
+     * Drain outbound UDP datagrams deferred while no operational LwIP netif was available.
+     * Must be called on the Matter chip thread. Safe to call when deferral is disabled (no-op).
+     */
+    static void FlushDeferredSendQueue();
```
jmartinez-silabs, line 69: doesn't seem to be called anywhere. Can we delete this?

src/inet/DeferredUdpSendQueueLwIP.cpp
```diff
@@ -0,0 +1,252 @@
...
+    SwapDeferredRings(gDeferredRing, pending);
+
+    while (!pending.Empty())
+    {
+        DeferredUdpSlot slot = pending.PopFront();
+
+        bool ready     = false;
+        err_t probeErr = EndPointStateLwIP::RunOnTCPIPRet([&]() -> err_t {
+            ready = IsOutboundNetifReadyForUdp(slot.pktInfo);
+            return ERR_OK;
```
jmartinez-silabs, line 186: Here you didn't address my comment when moving the stuff.

You check if it needs to be defered or not. If interface is up the you call Flush. and in Flush you check again if interface is up. 
```
    CHIP_ERROR deferProbeErr = DeferredUdpSendQueueLwIP::ProbeDefer(*pktInfo, shouldDefer);
    VerifyOrReturnError(deferProbeErr == CHIP_NO_ERROR, deferProbeErr);
    if (shouldDefer)
    {
        return DeferredUdpSendQueueLwIP::Enqueue(this, pktInfo, std::move(msg));
    }
    DeferredUdpSendQueueLwIP::Flush();
```

This is redundant or are you expecting the interface to possibly go down while flushing? I don't think we would turn off the interface while sending. If there is a Network issue then those deferred packets are probably not worth keeping either.. I don't know. please clarify

src/inet/UDPEndPointImplLwIP.h
jmartinez-silabs: Ok I understand why it is friend class. But could we not instead make make our own class with `DeferredUdpSendQueueLwIP` code in it and our class inherits `UDPEndPointImplLwIP`

name are irreleveant (you could think of a better name) here but just as an example
```
`class LowPowerUDPEndPointImplLwIP : UDPEndPointImplLwIP`
{
public:
   CHIP_ERROR PerformLwIPUdpSend(const IPPacketInfo * pktInfo, chip::System::PacketBufferHandle && msg);    
   CHIP_ERROR FlushOneDeferred(const IPPacketInfo * pktInfo, chip::System::PacketBufferHandle && msg);

    /**
     * If the netif is not ready for this destination, set outShouldDefer true so the send should be
     * queued. Returns a mapped CHIP error if the TCPIP thread bridge fails.
     */
    static CHIP_ERROR ProbeDefer(const IPPacketInfo & pktInfo, bool & outShouldDefer);

    static CHIP_ERROR Enqueue(UDPEndPointImplLwIP * self, const IPPacketInfo * pktInfo, System::PacketBufferHandle && msg);

    static void PurgeForEndpoint(UDPEndPointImplLwIP * self);

    static void Flush();

private:

    struct DeferredUdpSlot
    {
        UDPEndPointHandle epHandle;
        IPPacketInfo pktInfo;
        System::PacketBufferHandle msg;
    };

    struct DeferredUdpRing
   {
     ...
    }
    DeferredUdpRing mDeferredRing;
}
```

If this works, and i think it can with some tweak/fix, then we have so many added benefits.
We avoid modification to existing/common CSA UDPEndpointImpLwip.
If it receives updates/fix on CSA we would inherit them automatically.
We can extend the our class as we want with minimal impact with csa code.
It can be more flexible and robust.


Let me know if you see issues with this approach

chirag-silabs:
> This is a new feature. Your new class could be unit tested

Sure will add the unit test.

src/inet/DeferredUdpSendQueueLwIP.cpp
chirag-silabs: updated

src/inet/UDPEndPointImplLwIP.h
chirag-silabs: yes, I have thought of this but it is not possible due to the way Inet is written to be done cleanly
Currently, the UDPEndpoint Instances are allocted from `EndPointManagerImplPool<UDPEndPointImpl>`, where UDPEndPointImpl is the alias from the selected impl header. Then we need to remove https://github.com/project-chip/connectedhomeip/blob/master/src/inet/UDPEndPointImplLwIP.h#L83 and have it in the new class.

Plus the SendMsgImpl is the function which is customized the most which is private in the UDPEndpointLwIP so the changes will anywho it will be needed on the same file.

This changes anyhow doesn't need that much of refactor. Even when trying to inherit since the modification is in the UDPEndPointImplLwIP.cpp SendMsgImpl, again we need to include the class to use the PerformLwIPUdpSend or FlushOneDeferred from the subclass itself

src/inet/EndPointStateLwIP.h
chirag-silabs: Same as mentioned below, this is to use the protected functions in my class

src/inet/UDPEndPointImplLwIP.h
chirag-silabs: Plus having a couple of implementations as well for when queuing is needed and when not that will needed, i.e., it will be trhat the file DeferredUdpSendQueueLwIP should always be compile with the code.

src/inet/UDPEndPointImplLwIP.h
chirag-silabs: This will be used to call from the platform when we are connected to the AP

src/inet/DeferredUdpSendQueueLwIP.cpp
chirag-silabs: This check will be needed since ProbeDefer() is only for the current pktInfo. Flush() handles previously deferred entries that can target different interfaces/routes, so it is must validate readiness per slot. It also covers the netif state changes between calls (link/address/default route changes). I’ve added comments in UDPEndPointImplLwIP.cpp and DeferredUdpSendQueueLwIP.cpp to make this explicit.

src/inet/DeferredUdpSendQueueLwIP.cpp
chirag-silabs: added the unit test coverage

src/inet/BUILD.gn
```diff
@@ -50,6 +50,10 @@ buildconfig_header("inet_buildconfig") {
     "INET_CONFIG_ENABLE_TCP_ENDPOINT=${chip_inet_config_enable_tcp_endpoint}",
     "INET_CONFIG_ENABLE_UDP_ENDPOINT=${chip_inet_config_enable_udp_endpoint}",
     "HAVE_LWIP_RAW_BIND_NETIF=true",
+
+    # [SL-ONLY] Queuing the UDP packets for until the netif is up
```
jmartinez-silabs:
```suggestion
    # [SL-ONLY] Queuing the UDP packets until the netif is up
```

src/inet/BUILD.gn
```diff
@@ -184,6 +188,7 @@ static_library("inet") {
     #  - UDPEndPointImplSockets.h
     sources += [
       "BasicPacketFilters.h",
+      "DeferredUdpSendRing.h",
```
jmartinez-silabs, line 191: should that not be conditional addition on sl_inet_config_udp_lwip_queue_until_netif_ready == true?

src/inet/DeferredUdpSendQueueLwIP.cpp
jmartinez-silabs: ok

src/inet/BUILD.gn
```diff
@@ -193,6 +198,12 @@ static_library("inet") {
 
     if (!chip_system_config_use_network_framework) {
       sources += [ "UDPEndPointImpl${chip_system_config_inet}.cpp" ]
+      if (chip_system_config_inet == "LwIP") {
+        sources += [
+          "DeferredUdpSendQueueLwIP.cpp",
```
jmartinez-silabs, line 203: same here should it be conditional on sl_inet_config_udp_lwip_queue_until_netif_ready

jmartinez-silabs (approved):
Not a fan of the friend class requirement/usage with this approach. 
I believe you that my inheritance suggestion might not be suffisant but maybe there is other options to explore.

I'm approving as I don't have other suggestion but I would suggest to get this reviewed by other members of the team maybe they would have ideas. Maybe Louis-Philip or Ricardo would have some clever ideas.

src/inet/DeferredUdpSendRing.h
```diff
@@ -0,0 +1,103 @@
...
+    size_t count = 0;
+
+    bool Empty() const { return count == 0; }
+    bool IsFull() const { return count == Capacity; }
+    size_t ActiveCount() const { return count; }
+
+    // Enqueue at the logical tail (may drop oldest entry when full).
+    void PushBack(Slot && slot)
+    {
+        if (count == Capacity)
```
jmartinez-silabs:
```suggestion
        if (IsFull())
```

src/inet/DeferredUdpSendQueueLwIP.cpp
chirag-silabs: Yes it is thread safe
we use the same as the rest of the UDPEndpointImpl i.e., SendMsgImpl and CloseImpl which is already guarded with assertChipStackLockedByCurrentThread and with the flush as well, also added the assert on chip stack lock

src/inet/BUILD.gn
chirag-silabs: I tried to add it but the issue with this was the the unit test will not run for this file.
This is to ensure the unit test run on this file.

And since nothing include this other than DeferredUdpSendQueue.cpp under condition, keeping this out shouldn't cause any issues

src/inet/BUILD.gn
chirag-silabs: added the condition

chirag-silabs:
Merging to the feature branch, still open for suggestions :D

PR #932, opened 2026-04-23
[SL-ONLY] Modify Wake-on-Matter

jepenven-silabs:
#### Summary
Add the necessary code for NCP init, also add the necessary source for the phase 1 support.

the `iostream_uart.c` is copy pasted from the OT source code as it was easier to do than to demand a component refactor from the OT team. This fille will be deleted once we have CPC support.

#### Related issues
MATTER-5190

#### Testing
Tested with an OTBR

#### Readability checklist

code readable:

-   [ ] PR title is
-   [ ] Apply the
        rule (coding style)
-   [ ] PR size is short

See: [Pull Request Guidelines](https://project-chip.github.io/connectedhomeip-doc/contributing/pull_request_guidelines.html)

### DIFF
--- a/examples/wake-on-matter/silabs/src/AppTask.cpp
+++ b/examples/wake-on-matter/silabs/src/AppTask.cpp
@@ -342,9 +342,11 @@ void AppTask::ButtonEventHandler(uint8_t button, uint8_t btnAction)
 }
 
 // To prevent linkage failure
+#if SL_OPENTHREAD_MULTI_PAN_ENABLE
 extern "C" void otAppNcpInit(otInstance * aInstance);
 
 static otInstance * sInstance = NULL;
+#endif
 
 extern "C" void sl_ot_ncp_init(void)
 {
--- a/examples/wake-on-matter/silabs/src/AppTask.cpp
+++ b/examples/wake-on-matter/silabs/src/AppTask.cpp
@@ -340,3 +340,16 @@ void AppTask::ButtonEventHandler(uint8_t button, uint8_t btnAction)
     button_event.Handler            = BaseApplication::ButtonHandler;
     AppTask::GetAppTask().PostEvent(&button_event);
 }
+
+// To prevent linkage failure
+extern "C" void otAppNcpInit(otInstance * aInstance);
+
+static otInstance * sInstance = NULL;
+
+extern "C" void sl_ot_ncp_init(void)
+{
+#if SL_OPENTHREAD_MULTI_PAN_ENABLE
+    sInstance = otInstanceInitMultiple(1); // 1 NCP instance
+    otAppNcpInit(sInstance);
+#endif
+}
--- a/examples/wake-on-matter/silabs/src/AppTask.cpp
+++ b/examples/wake-on-matter/silabs/src/AppTask.cpp
@@ -351,7 +351,9 @@ static otInstance * sInstance = NULL;
 extern "C" void sl_ot_ncp_init(void)
 {
 #if SL_OPENTHREAD_MULTI_PAN_ENABLE
-    sInstance = otInstanceInitMultiple(1); // 1 NCP instance
+    // Matter Stack uses instances at index 0
+    // NCP instance will be at index 1
+    sInstance = otInstanceInitMultiple(1);
     otAppNcpInit(sInstance);
 #endif
 }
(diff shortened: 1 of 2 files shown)


examples/wake-on-matter/silabs/src/iostream_uart.c
```diff
@@ -0,0 +1,234 @@
+/*
```
jmartinez-silabs, line 1: NIT: maybe the file name could be renamed to explain its usage.
my understanding is that this is the uart stream for the OT NCP communication to host

jmartinez-silabs (approved):

examples/wake-on-matter/silabs/src/iostream_uart.c
jepenven-silabs: yeah, but this is also the name in the OT task, since it's a copy paste I want to preserve the name that way should the build system pick up both will have two object file with the same name which should trigger some warning...

examples/wake-on-matter/silabs/src/iostream_uart.c
```diff
@@ -0,0 +1,234 @@
...
+ *  Copyright (c) 2023, The OpenThread Authors.
```
lpbeliveau, line 2: Since this is a copy from an OT file, could we instead have this in the sdk-copies file in matter-support 
[here](https://github.com/SiliconLabsSoftware/matter_support/tree/main/sdk-copies)?

Since we already had an openthread folder in there it would belong.

examples/wake-on-matter/silabs/src/AppTask.cpp
```diff
@@ -340,3 +340,18 @@ void AppTask::ButtonEventHandler(uint8_t button, uint8_t btnAction)
...
+#if SL_OPENTHREAD_MULTI_PAN_ENABLE
+extern "C" void otAppNcpInit(otInstance * aInstance);
+
+static otInstance * sInstance = NULL;
+#endif
+
+extern "C" void sl_ot_ncp_init(void)
+{
+#if SL_OPENTHREAD_MULTI_PAN_ENABLE
+    sInstance = otInstanceInitMultiple(1); // 1 NCP instance
```
lpbeliveau: Could we use `otInstanceInitSingle()` instead?

examples/wake-on-matter/silabs/src/iostream_uart.c
jepenven-silabs: It will be remove next cycle, it's really a stop gap in order to release something in June. So the uglier it is, the less chance we forget to remove it

examples/wake-on-matter/silabs/src/AppTask.cpp
jepenven-silabs: No we could not, this only work with multiples OT instance as of now (one for NCP one for Matter). We might get away with using a single instance but that would most likely need a bit of refactoring on the OT stack itself. The `1` refers to the index of the instance we want to create in this case, not the number of instances itself.

sabollim-silabs: its better to have null check on sInstance

sabollim-silabs: NIT: the comment gives a wrong impression as 1 means only 1 instance. can we update comment to  reflect 1 is Index. if possible avoid magic number and use a variable like index etc

examples/wake-on-matter/silabs/src/AppTask.cpp
```diff
@@ -340,3 +340,18 @@ void AppTask::ButtonEventHandler(uint8_t button, uint8_t btnAction)
     button_event.Handler            = BaseApplication::ButtonHandler;
     AppTask::GetAppTask().PostEvent(&button_event);
 }
+
+// To prevent linkage failure
```
sabollim-silabs, line 344: can we expand on this comment, as i feel it doesnt convey complete meaning

examples/wake-on-matter/silabs/src/AppTask.cpp
```diff
@@ -340,3 +340,18 @@ void AppTask::ButtonEventHandler(uint8_t button, uint8_t btnAction)
     button_event.Handler            = BaseApplication::ButtonHandler;
     AppTask::GetAppTask().PostEvent(&button_event);
 }
+
```
sabollim-silabs, line 343: can we have a TODO which mentions this should be removed after the actual fix, so it can be tracked

examples/wake-on-matter/silabs/src/AppTask.cpp
jepenven-silabs: in a void function this won't do anything.

lpbeliveau-silabs (approved):

sabollim-silabs (approved):

PR #938, opened 2026-05-07
[SL-UP] Correcting the include file for the application_properties.h

chirag-silabs:
#### Summary
In OTAConfig.cpp, the include file is for the application_properties.h which is in the folder bootloader/platform/bootloader/api/application_properties.h

The component which is used to include this is https://github.com/SiliconLabsSoftware/sisdk-prerelease/blob/sisdk-2026.6/bootloader/platform/bootloader/component/bootloader_application_properties_header.slcc
Earlier it was working since the path for include was coming via some other components, but with the build 543 for series 3 it is not building anymore. 
Correcting the path so that it utilizes the correct include path for both Series 2 and Series 3

#### Related issues
NA

#### Testing
Local build test and CI

### DIFF
--- a/examples/platform/silabs/OTAConfig.cpp
+++ b/examples/platform/silabs/OTAConfig.cpp
@@ -22,7 +22,7 @@
 
 #ifndef SLI_SI91X_MCU_INTERFACE
 
-#include "application_properties.h"
+#include "api/application_properties.h"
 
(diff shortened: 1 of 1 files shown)


sabollim-silabs (approved):

rosahay-silabs (approved):
LGTM, but for CSA do change the include path in the GN build files to reflect this change, as I do suspect it should work without this change as well in CSA.

PR #976, opened 2026-05-26
[SL-UP]Add an API to check the Initialize state of AppTask

jmartinez-silabs:
#### Summary
Slight change in some refactor app lead cases where `MatterPostChangeAttributeCallback` was propagated in the AppTask before it is initialized leading to a crash.

For example setting StartUpOnOff attribute to any value could,(would in the case of Toggle) trigger a early OnOff state change.

Fix:
-Add a Boolean/Api to our BaseApplication, usable by the inheriting `AppTaskImpl`, to indicate that the AppTask is fully initialized.
- Set the bool, mIsApplicationInitialized, at the end of `BaseApplication::Init()` once `BaseInit()` and `AppInit()` are successfully called.
- Check IsApplicationInitialized() before propagating the MatterPostChangeAttributeCallback to te `AppTaskImpl`
 

#### Related issues
fixes https://jira.silabs.com/browse/MATTER-6412 and https://jira.silabs.com/browse/MATTER-6361

#### Testing
- Commission the cmp lighting-app.
- Set StartUpOnOff attribute to Toggle(2) : `mattertool onoff write start-up-on-off 2 $NODE_ID 1`
- Reboot the device, Led state Toggles every reboot, app doesn't crash anymore.
- run Test_TC_OO_2_4.yaml

### DIFF
--- a/examples/platform/silabs/BaseApplication.cpp
+++ b/examples/platform/silabs/BaseApplication.cpp
@@ -354,7 +354,7 @@ CHIP_ERROR BaseApplication::Init()
 #endif // SL_CATALOG_ZIGBEE_ZCL_FRAMEWORK_CORE_PRESENT
     SILABS_TRACE_END_ERROR(TimeTraceOperation::kAppInit, err);
     SILABS_TRACE_END_ERROR(TimeTraceOperation::kBootup, err);
-    mIsApplicationInitialized = true;
+    if (err == CHIP_NO_ERROR) { mIsApplicationInitialized = true };
     return err;
 }
 
--- a/examples/platform/silabs/BaseApplication.cpp
+++ b/examples/platform/silabs/BaseApplication.cpp
@@ -354,6 +354,7 @@ CHIP_ERROR BaseApplication::Init()
 #endif // SL_CATALOG_ZIGBEE_ZCL_FRAMEWORK_CORE_PRESENT
     SILABS_TRACE_END_ERROR(TimeTraceOperation::kAppInit, err);
     SILABS_TRACE_END_ERROR(TimeTraceOperation::kBootup, err);
+    mIsApplicationInitialized = true;
     return err;
 }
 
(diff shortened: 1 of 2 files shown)


Sarthak-Shaha (approved):

examples/platform/silabs/BaseApplication.h
```diff
@@ -118,6 +118,13 @@ class BaseApplication
      */
     void UnlinkAppLed() { sAppActionLed = nullptr; }
 
+    /**
+     * @brief Check if the application is initialized
+     *
+     * @return Set to true when Init() was called successfully
```
lpbeliveau, line 124: Currently it returns to rue no matter what the Init result is.

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -354,6 +354,7 @@ CHIP_ERROR BaseApplication::Init()
 #endif // SL_CATALOG_ZIGBEE_ZCL_FRAMEWORK_CORE_PRESENT
     SILABS_TRACE_END_ERROR(TimeTraceOperation::kAppInit, err);
     SILABS_TRACE_END_ERROR(TimeTraceOperation::kBootup, err);
+    mIsApplicationInitialized = true;
```
lpbeliveau: I know we have a early return whenever `err != CHIP_NO_ERROR`, but for the sake of consistency could we have a ternary here?

Like:
`mIsApplicationInitialized = (err = CHIP_NO_ERROR) ? true : false;`

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -1128,13 +1129,16 @@ bool BaseApplication::GetProvisionStatus()
...
-    CustomerAppTask::GetAppTask().DMPostAttributeChangeCallback(attributePath, type, size, value);
 #ifdef SL_CATALOG_ZIGBEE_ZCL_FRAMEWORK_CORE_PRESENT
     EndpointId endpointId   = attributePath.mEndpointId;
     ClusterId clusterId     = attributePath.mClusterId;
     AttributeId attributeId = attributePath.mAttributeId;
     MultiProtocolDataModel::WriteMatterAttributeValueToZigbee(endpointId, clusterId, attributeId, value, type);
 #endif // SL_CATALOG_ZIGBEE_ZCL_FRAMEWORK_CORE_PRESENT
+
+    // Verify that the App layer is initialized before propagating attribute changes callback to it.
+    VerifyOrReturn(CustomerAppTask::GetAppTask().IsApplicationInitialized());
```
lpbeliveau, line 1140: Should we have this check at the method entrance? Or is there a benefit to updating only Zigbee on our CMP?

lpbeliveau (approved):

miduggan24 (approved):

examples/platform/silabs/BaseApplication.h
jmartinez-silabs: that is not true. If we have errors we early return

examples/platform/silabs/BaseApplication.cpp
jmartinez-silabs: I don't think i need to set it to false, and we should not reach this if a error preventing baseApp or AppTask to run occured.
Worst case I can do
if (err == CHIP_NO_ERROR) { mIsApplicationInitialized = true };

examples/platform/silabs/BaseApplication.cpp
jmartinez-silabs: There is some (limited) benefit to write to ZB datamodel early.

examples/platform/silabs/BaseApplication.cpp
jmartinez-silabs:
```suggestion
    if (err == CHIP_NO_ERROR) { mIsApplicationInitialized = true };
```

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -354,6 +354,7 @@ CHIP_ERROR BaseApplication::Init()
 #endif // SL_CATALOG_ZIGBEE_ZCL_FRAMEWORK_CORE_PRESENT
     SILABS_TRACE_END_ERROR(TimeTraceOperation::kAppInit, err);
     SILABS_TRACE_END_ERROR(TimeTraceOperation::kBootup, err);
+    if (err == CHIP_NO_ERROR) { mIsApplicationInitialized = true };
```
jmartinez-silabs:
```suggestion
    if (err == CHIP_NO_ERROR) { mIsApplicationInitialized = true; }
```

PR #993, opened 2026-06-03
[SL-UP] Add memory CLI api

jepenven-silabs:
#### Summary

to get heap usage without Diagnostic cluster (code size savings)

#### Testing
Tested on MG24 and series 3

#### Readability checklist

code readable:

-   [ ] PR title is
-   [ ] Apply the
        rule (coding style)
-   [ ] PR size is short

See:

### DIFF
--- a/examples/platform/silabs/MatterShell.cpp
+++ b/examples/platform/silabs/MatterShell.cpp
@@ -115,6 +115,43 @@ void cmdSilabsInit()
 
 #endif // SL_CATALOG_CLI_PRESENT
 
+#include "sl_memory_manager.h"
+
+namespace MemoryShellCommands {
+
+Engine sShellMemorySubCommands;
+
+CHIP_ERROR MemoryCommandHandler(int argc, char ** argv)
+{
+    if (argc == 0)
+    {
+        sShellMemorySubCommands.ForEachCommand(Shell::PrintCommandHelp, nullptr);
+        return CHIP_NO_ERROR;
+    }
+    return sShellMemorySubCommands.ExecCommand(argc, argv);
+}
+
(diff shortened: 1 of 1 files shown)


examples/platform/silabs/MatterShell.cpp
```diff
@@ -115,6 +115,43 @@ void cmdSilabsInit()
 
 #endif // SL_CATALOG_CLI_PRESENT
 
+#include "sl_memory_manager.h"
```
Sarthak-Shaha, line 118: Would it be better to move it along with other includes?

jmartinez-silabs (approved):

examples/platform/silabs/MatterShell.cpp
```diff
@@ -115,6 +115,43 @@ void cmdSilabsInit()
...
+
+namespace MemoryShellCommands {
```
lpbeliveau, line 120: Being Mr. CodeSize I suppose you already thought about that but I'll still ask, do we want some gating for this feature, say we want the shell but not the heap display? 

Also, why not have a memory subdir like the icd, im, ble and tracing shell commands?

rcasallas-silabs (approved):

lpbeliveau (approved):

examples/platform/silabs/MatterShell.cpp
jepenven-silabs: In that case API are already present by default since we use the memory manager by default. Removing the shell completely is the only way of having some gain here..

examples/platform/silabs/MatterShell.cpp
jepenven-silabs: should be good for now

PR #1002, opened 2026-06-09
[SL-ONLY] Propagate binding commands to rangehood on oven-reboot

arun-silabs:
#### Summary

Fixes oven-to-rangehood binding behavior on oven reboot, rangehood reboot, and rangehood state persistence across reboot.

Oven app

- After reboot, CookTop and CookSurface OnOff are forced Off at startup (safety). Binding commands are not sent until network connectivity is available (IPv4 or IPv6), then OnOff and FanControl bindings are pushed to the rangehood.
- Before each binding propagation, cached CASE sessions to bound peers are expired (ExpireAllSessions + ReleaseSession) so a fresh CASE handshake is used instead of reusing a stale session after a rangehood reboot.
- NotifyBoundClusterChanged is deferred via ScheduleWork so binding/CASE work runs on the next Matter event-loop turn, avoiding failures from nested platform or attribute callbacks.

Rangehood app

- Changes OnOff (EP2) and FanMode (EP1) storage from RAM to NVM in both Wi-Fi and Thread data models, so light and fan state survive reboot.

#### Related issues

[MATTER-6677](https://jira.silabs.com/browse/MATTER-6677)

#### Testing

- Build and flash oven app (matter_wifi_soc_oven_app_freertos, brd4338a) and rangehood app with updated data model
- Commission both devices on the same fabric; configure CookTop bindings to rangehood OnOff (EP2) and FanControl (EP1)
- Turn CookTop ON - verify rangehood light turns on and fan activates
- Turn CookTop OFF - verify rangehood light and fan turn off
- Reboot oven with rangehood left ON - after Wi-Fi connects, verify logs show binding sync and rangehood turns off:
  TryPendingBindingSync: propagating CookTop Off to bound peers (or equivalent)
  CookTop OnOff bound unicast command success
  CookTop FanControl FanMode bound write success
- Reboot rangehood - verify OnOff and FanMode retain last values (NVM persistence)

#### Readability checklist

code readable:

-   [x] PR title is
-   [x] Apply the
        rule (coding style)
-   [x] PR size is short
-   [x] Try to avoid "squashing" and "force-update" in commit history
-   [x] CI time didn't increase

See:


rosahay-silabs (approved):

examples/oven-app/silabs/src/OvenBindingHandler.cpp
```diff
@@ -124,3 +124,27 @@ CHIP_ERROR CookTopBindingTrigger(CookTopBindingContext * context)
                         ChipLogError(AppServer, "CookTopBindingTrigger: null context"));
     return DeviceLayer::PlatformMgr().ScheduleWork(TriggerBindingWork, reinterpret_cast<intptr_t>(context));
 }
+
+void CookTopBindingPropagateState(EndpointId cookTopEndpoint, bool cookTopOn)
+{
+    for (ClusterId clusterId : { OnOff::Id, FanControl::Id })
+    {
+        CookTopBindingContext * context = Platform::New<CookTopBindingContext>();
```
jmartinez-silabs, line 141: why does it need to be an dynamic allocation. You delete it at the end of this loop

jmartinez-silabs (commented):
I am questionning the logic of `CookTopBindingPropagateState`

There a is a deep call stack created for apparently no reason.

Both calls of CookTopBindingPropagateState comes from the matter Task context.

`CookTopBindingPropagateState` calls `CookTopBindingTrigger` which, from my understanding, is just used to call ScheduleWork so the call is done on the matter task context (which we already are)

and the scheduleWork is used to call `TriggerBindingWork` which itself calls `NotifyBoundClusterChanged`. This means we have 


CookTopBindingPropagateState
-> CookTopBindingTrigger
--> ScheduleWork 
---> TriggerBindingWork
----> NotifyBoundClusterChanged

All this could be simplified to a single function that calls `NotifyBoundClusterChanged`

arun-silabs:
> I am questionning the logic of `CookTopBindingPropagateState`
> 
> There a is a deep call stack created for apparently no reason.
> 
> Both calls of CookTopBindingPropagateState comes from the matter Task context.
> 
> `CookTopBindingPropagateState` calls `CookTopBindingTrigger` which, from my understanding, is just used to call ScheduleWork so the call is done on the matter task context (which we already are)
> 
> and the scheduleWork is used to call `TriggerBindingWork` which itself calls `NotifyBoundClusterChanged`. This means we have
> 
> CookTopBindingPropagateState -> CookTopBindingTrigger --> ScheduleWork ---> TriggerBindingWork ----> NotifyBoundClusterChanged
> 
> All this could be simplified to a single function that calls `NotifyBoundClusterChanged`

Hmm yes...
The extra ScheduleWork layer isn't needed here. I originally added it because I wanted to be safe against deadlocks if a future caller invoked CookTopBindingPropagateState from a different thread. But yeah, we can add a comment to the function to make sure that it has to be called from Matter Task context only. I've made changes as suggested

examples/oven-app/silabs/src/OvenBindingHandler.cpp
arun-silabs: I placed delete logic at the wrong place. It should only be deleted on immediate failure. On success, BindingManager keeps the context until CASE completes and ContextReleaseHandler runs, that can be after CookTopBindingPropagateState returns since the operation is async. So i think the dynamic allocation is needed here.

jmartinez-silabs: Ok that make sense then now that the delete is moved to only the failure case.

jmartinez-silabs (approved):

examples/oven-app/silabs/src/OvenBindingHandler.cpp
```diff
@@ -19,6 +20,26 @@ using namespace chip::app::Clusters;
 
 namespace {
 
+// Drop cached CASE sessions to bound peers so BindingManager re-handshakes after a peer reboot.
```
sabollim-silabs:
```suggestion
// Drop cached CASE sessions to bound peers so BindingManager re-handshakes after every binding action
```

examples/oven-app/silabs/src/OvenBindingHandler.cpp
```diff
@@ -118,9 +139,30 @@ CHIP_ERROR InitOvenBindingHandler()
     return DeviceLayer::PlatformMgr().ScheduleWork(InitBindingMgrWork);
 }
 
-CHIP_ERROR CookTopBindingTrigger(CookTopBindingContext * context)
+void CookTopBindingPropagateState(EndpointId cookTopEndpoint, bool cookTopOn)
 {
-    VerifyOrReturnError(context != nullptr, CHIP_ERROR_INVALID_ARGUMENT,
-                        ChipLogError(AppServer, "CookTopBindingTrigger: null context"));
-    return DeviceLayer::PlatformMgr().ScheduleWork(TriggerBindingWork, reinterpret_cast<intptr_t>(context));
+    ReleaseBoundPeerSessions(cookTopEndpoint);
```
sabollim-silabs, line 137: add why we need to remove session every time

sabollim-silabs (approved):

examples/oven-app/silabs/src/OvenBindingHandler.cpp
arun-silabs: added a comment

PR #1015, opened 2026-06-12
[SL-ONLY] Set ZLL not factory new for both Sequential and Concurrent CMP after matter commissioning completes

jmartinez-silabs:
### Summary
SQA reported instances failure to rejoin Thread/Matter network after a powercycle.
Log show that ZLL Init/Touchlink are active at boot even when we are part of a Matter network. This should not be the case after [#255](https://github.com/SiliconLabsSoftware/matter_sdk/pull/255) from previous releases addressing exactly this.

ZB Touchlink announcement have been proven in the pass to affect Thread attachment and SRP registration at bootup.

After investigation, we found that the Zigbee stack’s leave-network sequence clears our Not FactoryNew setting. On a Leave Network request, the stack runs a Leave Announcement period before the device actually leaves. If the device leaves without communicating with another device during that period, the stack resets the FactoryNew token and overwrites our setting.

The following changes ensure that this sequence do not impact our desired setting anymore.
- Add a Delay to our ZLLNotFactoryNew call: 
- Clear the ZB FactoryNew state for both Sequential and Concurrent.

### Related issues
https://jira.silabs.com/browse/MATTER-6389

### Testing
Commission CMP sequential app on Matter but not on Zigbee. Reboot the device. 
 - Validate that ZLL Init/touchlink is not done at bootup.

### DIFF
--- a/examples/platform/silabs/BaseApplication.cpp
+++ b/examples/platform/silabs/BaseApplication.cpp
@@ -175,6 +175,8 @@ bool sHaveBLEConnections = false;
 
 constexpr uint32_t kLightTimerPeriod = static_cast<uint32_t>(pdMS_TO_TICKS(10));
 
+constexpr System::Clock::Milliseconds32 kZbLeaveAnnouceDelay = System::Clock::Milliseconds32(1000);
+
 uint8_t sAppEventQueueBuffer[APP_EVENT_QUEUE_SIZE * sizeof(AppEvent)];
 osMessageQueue_t sAppEventQueueStruct;
 constexpr osMessageQueueAttr_t appEventQueueAttr = { .cb_mem  = &sAppEventQueueStruct,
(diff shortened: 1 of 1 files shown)


Sarthak-Shaha (approved):

miduggan24 (approved):

examples/platform/silabs/BaseApplication.cpp
selissia, line 352: Should we also do  Zigbee::ZLLNotFactoryNew(); here?

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -1107,8 +1110,10 @@ void BaseApplication::OnPlatformEvent(const ChipDeviceEvent * event, intptr_t)
 #ifdef SL_CATALOG_ZIGBEE_STACK_COMMON_PRESENT
 #ifdef SL_MATTER_ZIGBEE_SEQUENTIAL // Matter Zigbee sequential
         Zigbee::RequestLeave();
-        Zigbee::ZLLNotFactoryNew();
 #endif // SL_MATTER_ZIGBEE_SEQUENTIAL
+        constexpr System::Clock::Milliseconds32 kZbLeaveAnnouceDelay = System::Clock::Milliseconds32(1000);
```
selissia: Should this be under #ifdef SL_MATTER_ZIGBEE_SEQUENTIAL like Zigbee::ZLLNotFactoryNew(); was before?

selissia: Ah, never mind. The PR overview says this is intentional.

selissia (approved):

examples/platform/silabs/BaseApplication.cpp
jmartinez-silabs: I think it is a safe assumption. I will add it.

PR #1017, opened 2026-06-15
[SL-UP] Updated virtual function SleepButtonActionHandler to non-virtual

silabs-ArghyaD:
### Summary
Made SleepButtonActionHandler a non-virtual function as the virtual API is causing the device to get stuck during OTA after soft reset. Similar to https://github.com/SiliconLabsSoftware/matter_sdk/pull/819

### Related issues
MATTER-6373

### Testing
- Tested OTA for ~10 times, the issue was not seen.
- Tested BTN0 functionality.

### DIFF
--- a/src/platform/silabs/platformAbstraction/SilabsPlatform.h
+++ b/src/platform/silabs/platformAbstraction/SilabsPlatform.h
@@ -55,7 +55,7 @@ public:
     static SilabsButtonCb mButtonCallback;
     uint8_t GetButtonState(uint8_t button) override;
 #if defined(SL_ICD_ENABLED) && SL_ICD_ENABLED == 1
-    void SleepButtonActionHandler(void) override;
+    void SleepButtonActionHandler(void);
 #endif // defined(SL_ICD_ENABLED) && SL_ICD_ENABLED == 1
 #endif // defined(SL_CATALOG_SIMPLE_BUTTON_PRESENT)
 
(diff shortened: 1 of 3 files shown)


jmartinez-silabs (commented):
IMO reverting this makes the code worst. 

What is the issue with having a virtual function for OTA?

chirag-silabs:
> IMO reverting this makes the code worst.
> 
> What is the issue with having a virtual function for OTA?

@jmartinez-silabs 
We have seen this earlier as well where the virtual function in the idle task is just not working as expected when OTA update is done and image is booting with the new image. The same was seen when the virtual watchdog timer was added. 
Not sure if vtable is getting corrupted/not populated with the virtual function and and on runtime its losing reference with the new image.

Will confirm with the platform team but this was the quick fix which can be done, move the API's to non virtual functions.

jmartinez-silabs:
@chirag-silabs @silabs-ArghyaD 
Yes, I understand that you had similar issue with the linked pr in the description. However this mean this is the second time we do "workaround" instead of understanding the root cause and fix it. 
This seem really odd that a virtual function would cause issue. This is a base feature of C++ this should just work. This should be fully investigated and understood it isn't normal. We could have bigger underlying issues if the linker doesn't to the right things.


Additionally, we are reverting clearner code as a workaround, but shouldn't always defining the function in SPAM, even if it doesn't do anything in some configuration be clearner, and fast to do that going back to the older code? What that evaluated/tested?

silabs-ArghyaD:
Updated the PR title and description. Kept the changes to as minimal as possible.

jmartinez-silabs (approved):

rosahay-silabs (approved):
LGTM, can we have a task to investigate this behavior of `vTables` not getting populated.

PR #1025, opened 2026-06-18
[SL-UP] MATTER-6705: Fix CMP Sequential chipDie

miduggan24:
### Summary
In https://github.com/SiliconLabsSoftware/matter_sdk/pull/1015, we introduce some changes to add delay to ZLLNotFactoryNewCall through StartTimer() in BaseApplication::Init() that runs without locking chip stack, which causes chipDie on assertion on app boot. 

Wrap the StartTimer() call in LockChipStack()/UnlockChipStack() 

### Related issues
MATTER-6705

### Testing
Build locally with lock, commission, reboot, verify no chipDie

### DIFF
--- a/examples/platform/silabs/BaseApplication.cpp
+++ b/examples/platform/silabs/BaseApplication.cpp
@@ -350,8 +350,10 @@ CHIP_ERROR BaseApplication::Init()
     if (nbOfMatterFabric != 0)
     {
         Zigbee::RequestLeave();
+        PlatformMgr().LockChipStack();
         RETURN_SAFELY_IGNORED DeviceLayer::SystemLayer().StartTimer(
             kZbLeaveAnnouceDelay, [](System::Layer *, void *) { Zigbee::ZLLNotFactoryNew(); }, nullptr);
+        PlatformMgr().UnlockChipStack();
     }
     else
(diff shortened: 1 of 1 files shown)


Sarthak-Shaha (approved):

jepenven-silabs (approved):

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -350,8 +350,10 @@ CHIP_ERROR BaseApplication::Init()
     if (nbOfMatterFabric != 0)
     {
         Zigbee::RequestLeave();
+        PlatformMgr().LockChipStack();
```
lpbeliveau, line 353: rather than add 2 new lock -> unlock, why not just take `PlatformMgr().UnlockChipStack();` at line 349 and move it after the if else block?

miduggan24: good point, let me test this to be certain

miduggan24: seems like we would need to either keep two lock/unlock, or change else -> if (nbOfMatterFabric == 0) and place unlock before. if we place after the else, the unlock will become the else body and only execute then. imo having two lock/unlock is cleaner than explicitly checking against this if statement

selissia (approved):

PR #1053, opened 2026-07-06
[Cherry-pick] Add memory CLI api (#993)

jepenven-silabs:
Add memory CLI api and Watchdog

### Testing
released


jmartinez-silabs:
This was Tagged as SL-UP in release_2.9. It should be open on CSA no? If you want to keep it only in matter_sdk then tag it SL-ONLY

jepenven-silabs:
> This was Tagged as SL-UP in release_2.9. It should be open on CSA no? If you want to keep it only in matter_sdk then tag it SL-ONLY

oups

PR #1073, opened 2026-07-13
[SL-ONLY]Improve retrieve_sl_commits.py to filter commits already present on target branch

jmartinez-silabs:
### Summary
retrieve_sl_commits.py is used to flag all commits with ou prefix labels ( SL-UP,SL-ONLY,SL-TEMP,CSA-CP), between two giving commits.

However during release_2.9.-1.6 developement cycle, some SL-ONLY and SL-TEMP were done on main branch and main branch was then merge multiple times into `release_2.9.-1.6`. With this situation the current script was flagging commit that were already present or already cherry-picked in `main` branch.

This PR adds two option: 
--target_branch,-t <branch>
--label, -l

When --target-branch is provided, each identified commit is checked against the provided target branch. If the commit is present on the branch the commit will be excluded from the output.
 
 If --label is also provided, then every commit will be outputted but with an additional annotation:
```
      [PRESENT on <branch>]       the exact SHA is reachable from the branch (git branch --contains).
      [CHERRY-PICKED on <branch>] the SHA is not on the branch, but a commit with the same title is.
      [MISSING from <branch>]     neither the SHA nor the title was found on the branch.
```
This helps identify, for example, which [SL-ONLY] commits already exist on the main branch.

example of of usage:
`python3 scripts/tools/silabs/retrieve_sl_commits.py ebc67e7f2c release_2.9-1.6 -t main -l`

### Related issues
N/A

### Testing
Manually tested by running the command locally and validation the output between release_2.9-1.6 and main.

### DIFF
--- a/scripts/tools/silabs/retrieve_sl_commits.py
+++ b/scripts/tools/silabs/retrieve_sl_commits.py
@@ -3,12 +3,15 @@ import argparse
 import subprocess
 from argparse import RawTextHelpFormatter
 
+# field separator for the git log output
+FIELD_SEP = "  |  "
+
 
 def get_git_log(start_sha, end_sha, prefixes):
     try:
         # Run the git log command with output format <Commit short hash>  |  <Author>  |  <Title>
         result = subprocess.run(
-            ['git', 'log', '--pretty=format:%h  |  %an  |  %s', f'{start_sha}..{end_sha}'],
+            ['git', 'log', f'--pretty=format:%h{FIELD_SEP}%an{FIELD_SEP}%s', f'{start_sha}..{end_sha}'],
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
             check=True,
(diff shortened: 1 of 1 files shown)


scripts/tools/silabs/retrieve_sl_commits.py
```diff
@@ -35,13 +38,93 @@ def get_git_log(start_sha, end_sha, prefixes):
...
+
+
 def main():
     parser = argparse.ArgumentParser(formatter_class=RawTextHelpFormatter, description="""
     This script will parse git logs for our silabs prefixes ([SL-UP], [SL-TEMP], [SL-ONLY] or [CSA-CP]) between the commit SHAs provided in parameters
     on the current git branch.
     It will then output, per prefix, the commit sha and commit Title in the following format)
     [PREFIX] commits:
     <full_commit_sha> -- <Commit_Title>
+
```
Sarthak-Shaha, line 122: Maybe this can be part of top level comment for the file?

Sarthak-Shaha (approved):

scripts/tools/silabs/retrieve_sl_commits.py
jmartinez-silabs: This is the help output. Shouldn't this description be part of the help?

scripts/tools/silabs/retrieve_sl_commits.py
```diff
@@ -35,13 +38,93 @@ def get_git_log(start_sha, end_sha, prefixes):
...
+    """
+    try:
+        result = subprocess.run(
+            ['git', 'log', target_branch, '--fixed-strings', f'--grep={subject}', '--pretty=format:%H'],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            check=True,
+            text=True
+        )
+        return bool(result.stdout.strip())
```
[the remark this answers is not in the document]
jmartinez-silabs: Acceptable risk as it would have to contain the full subject with the prefix [SL-AAAA]

miduggan24 (approved):

PR #1079, opened 2026-07-16
[SL-ONLY] Fan control cascading updates fix

arun-silabs:
### Summary

This PR fixes the cascading attribute handling in Fan Control app.

The app was remapping FanMode, PercentSetting, and SpeedSetting in ways that fought the cluster’s own cascade logic, causing write/read mismatches and missing subscription reports.

**Problem**
- Writing PercentSetting=1 was read back as 10 because a speed→fan-mode→percent remap overwrote small percent values.
- FanMode descending to Off produced 3 FanMode reports but only 2 PercentSetting reports (3 != 2), failing the report-count check.

**Fix**
Refactor FanControlManager to support feature-gated, directional synchronization between FanMode, SpeedSetting, and PercentSetting.
- MultiSpeed enabled
    - Direct FanMode writes now update SpeedSetting only.
    - The Fan Control cluster derives PercentSetting from SpeedSetting.
    - SpeedSetting changes synchronize only FanMode and SpeedCurrent. They no longer remap values back to other settings, preventing circular updates.
    - Direct PercentSetting writes update only PercentCurrent.

- MultiSpeed disabled
    - Direct FanMode writes map to PercentSetting using percentage bands derived dynamically from the current SpeedMax value via SpeedToPercent().
    - This replaces the previous fixed thresholds (1–33%, 34–66%, and 67–100%), allowing the mapping to adapt to the configured maximum speed.

- Additional changes
    - SetSpeedSetting() now uses the synchronous Attributes::SpeedSetting::Set() API.
    - This ensures the application delegate is invoked before the cluster's FanMode callback, allowing the expected PercentSetting subscription report to be emitted when transitioning to Off.

### Related issues

[MATTER-6315](https://jira.silabs.com/browse/MATTER-6315)

### Testing

Build and flash fan-control app on BRD4338A.
Run TC-FAN-3.1 (TC_FAN_3_1.py) and ensure that the test-case passes.

### DIFF
--- a/examples/fan-control-app/silabs/include/FanControlManager.h
+++ b/examples/fan-control-app/silabs/include/FanControlManager.h
@@ -51,12 +51,6 @@ public:
 
     void PercentSettingWriteCallback(uint8_t aNewPercentSetting);
     void SpeedSettingWriteCallback(uint8_t aNewSpeedSetting);
-    void FanModeWriteCallback(FanModeEnum aNewFanMode);
-
-    void SetPercentSetting(Percent aNewPercentSetting);
-    DataModel::Nullable<uint8_t> GetSpeedSetting();
-    DataModel::Nullable<Percent> GetPercentSetting();
-    void UpdateFanMode();
 
     static void UpdateClusterState(intptr_t arg);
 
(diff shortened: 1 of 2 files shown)


miduggan24 (commented):
can this PR be targeted to release branch? adding and removing methods like this will have implications on project upgrades.

plus it seems like based on ticket failure is seen on release branch and fix version is for the patch.

arun-silabs:
> can this PR be targeted to release branch? adding and removing methods like this will have implications on project upgrades.
> 
> plus it seems like based on ticket failure is seen on release branch and fix version is for the patch.

Created a new PR targeting release branch : https://github.com/SiliconLabsSoftware/matter_sdk/pull/1081

PR #1123, opened 2026-08-03
WIP: Add NWP HTTPS client demo and PAHO MQTTS client demo

rosahay-silabs:
Start http_client_demo after connectivity when services are enabled; turn on SiWx HTTP_CLIENT feature bit.


src/platform/silabs/services/cacert.pem.h
rosahay-silabs, line 1: THIS IS A SAMPLE CA CERT GENERATED LOCALLY

PR #1138, opened 2026-08-10
[SL-ONLY] Place LwIP memp pools in PSRAM when SiWx code classifiers is enabled

chirag-silabs:
### Summary
- Place Silabs LwIP `memp_memory_*` pools into PSRAM via `SL_SI91X_CODE_CLASSIFY(bss_to_psram, ..., SL_CODE_CLASS_FORCE_PSRAM)` when both `SL_SI91X_CODE_CLASSIFIER_ENABLE` and `SLI_SI91X_MCU_PSRAM_PRESENT` are defined.
- Keep the existing `.bss.lwip_*` section placement for builds without the PSRAM code classifier, so non-PSRAM / non-classifier platforms are unchanged.
- This will be used to testing the BSS buffers in the PSRAM by the SQA

### Related issues
NA

### Testing
Tested local builds and will be tested on matter_extension CI

### DIFF
--- a/src/lwip/silabs/arch/cc.h
+++ b/src/lwip/silabs/arch/cc.h
@@ -67,7 +67,39 @@ extern void LwIPLog(const char * fmt, ...);
     } while (0)
 
 // Place LwIP pools into their own subsections of .bss to make it easier to see
-// their sizes in the linker map file.
+// their sizes in the linker map file. On SiWx with the PSRAM code classifier,
+// place the pools in PSRAM via SL_SI91X_CODE_CLASSIFY.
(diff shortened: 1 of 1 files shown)


jmartinez-silabs (commented):
why is this sl-only?

chirag-silabs:
> why is this sl-only?

I was thinking about it, being SL-UP. Could use suggestions as well. It is needed for the SQA tests. Should I create it for CSA?

chirag-silabs:
Discussed with the team, pushing this change to CSA is valid.
Created a CSA PR: https://github.com/project-chip/connectedhomeip/pull/73496/changes

Hence closing this PR

PR #1141, opened 2026-08-11
[SL-ONLY] MATTER-6525: Adds PAHO MQTTs support

rosahay-silabs:
### Summary

- Add Paho MQTT/MQTTS client demo (`mqtt_client.c`) and start it from `BaseApplication` when `SL_MATTER_ENABLE_MQTT_SERVICE` is enabled after connectivity is up
- Enable LwIP sockets for dual-stack Paho (`SLI_PAHO_MQTT_OVER_LWIP`), raise FreeRTOS heap for host LwIP TLS, and fix static TCPIP mbox reuse so netconn recvmboxes do not reinit the tcpip queue
- Route Si91x heap diagnostics through `sl_si91x_memory_manager` and initialize PSA Crypto in `WifiInterfaceImpl` when `MBEDTLS_PSA_CRYPTO_C` is enabled

### Related issues

MATTER-6525

### Testing

Tested with locally hosted MQTT broker with TLS and CA certificate.


examples/platform/silabs/freertos_config/FreeRTOSConfig.h
```diff
@@ -282,6 +282,14 @@ See http://www.FreeRTOS.org/RTOS-Cortex-M3-M4.html. */
 #define EXTRA_HEAP_k 0
 #endif
 
+#ifdef MQTT_USE_HOST_LWIP_TLS
+/* Pays for mqtt_lwip stack 3->8KB and keeps ~10KB+ free for PSA RSA-2048.
+ * RAM headroom only allows ~+20KB total before linker assert. */
+#define MQTT_EXTRA_HEAP_k 28
```
sabollim-silabs, line 286: from above 20kb hardlimit comment . i assume we cannot allocate more than 20kb extra heap. but we are allocating 28k. am i missing anything here ?

examples/platform/silabs/freertos_config/FreeRTOSConfig.h
```diff
@@ -291,7 +299,7 @@ See http://www.FreeRTOS.org/RTOS-Cortex-M3-M4.html. */
 #define configTOTAL_HEAP_SIZE ((size_t) ((68 + EXTRA_HEAP_k) * 1024))
 #endif // SLI_SI91X_MCU_INTERFACE
 #else
-#define configTOTAL_HEAP_SIZE ((size_t) ((42 + EXTRA_HEAP_k) * 1024))
+#define configTOTAL_HEAP_SIZE ((size_t) ((72 + EXTRA_HEAP_k + MQTT_EXTRA_HEAP_k) * 1024))
```
sabollim-silabs, line 303: lets add a comment for why we inc to 72k , so it will be as a reference

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -1044,9 +1048,14 @@ void InitMatterServicesHandler(System::Layer * systemLayer, void * appState)
 #ifdef SL_MATTER_ENABLE_HTTP_SERVICE
     VerifyOrReturn(SL_STATUS_OK == http_client_demo_start(), ChipLogError(AppServer, "http_client_demo_start failed"));
 #endif // SL_MATTER_ENABLE_HTTP_SERVICE
+
+#ifdef SL_MATTER_ENABLE_MQTT_SERVICE
```
sabollim-silabs: why the ifdef on line 1052 varies with 98

sabollim-silabs (commented):
LGTM. skipped mqttclient.c as its same as wifi example. gave minor comments

examples/platform/silabs/MatterShell.cpp
```diff
@@ -102,7 +102,11 @@ void cmdSilabsInit()
 
 #endif // SL_CATALOG_CLI_PRESENT
 
+#if defined(SLI_SI91X_MCU_INTERFACE) && SLI_SI91X_MCU_INTERFACE
+#include "sl_si91x_memory_manager.h"
```
agarwal57, line 107: - Please add a comment here for this change. Is this related to Heap 4 and Heap 3?

src/lwip/freertos/sys_arch.c
```diff
@@ -165,7 +166,19 @@ err_t sys_mbox_new(sys_mbox_t * mbox, int size)
     }
 
 #if LWIP_FREERTOS_USE_STATIC_TCPIP_QUEUE
-    *mbox = xQueueCreateStatic((UBaseType_t) size, (UBaseType_t) sizeof(void *), gTCPIPMsgQueueStorage, &gTCPIPMsgQueue);
+    /* The static storage backs the tcpip thread's mbox only. Other mboxes (notably a
```
agarwal57: - i thought we said we will just go with the dynamic and not static. Are those changes in the next PR of this?

src/lwip/silabs/lwipopts-siwx.h
```diff
@@ -71,7 +71,13 @@
 #define ARP_QUEUEING (0)
 #define TCPIP_THREAD_NAME "LWIP"
 
+#if defined(SLI_PAHO_MQTT_OVER_LWIP) && SLI_PAHO_MQTT_OVER_LWIP
```
agarwal57: - is this the new macro we are using or picking from Wiseconnect? I am just trying to see if we can just remove Paho in the name as we do not intend to support multiple variants of MQTT.

src/platform/silabs/services/mqtt_client.c
```diff
@@ -0,0 +1,343 @@
+/***************************************************************************/
+/**
+ * @file mqtt_client.c
+ * @brief Paho MQTT/MQTTS client demo for Matter dual-stack (NWP sockets + host LwIP)
+ *******************************************************************************
+ * # License
+ * <b>Copyright 2025 Silicon Laboratories Inc. www.silabs.com</b>
```
agarwal57: - update the copyright year. This is generic comment and please ensure that this is correct in all the new files that are being added.

src/platform/silabs/services/mqtt_client.c
```diff
@@ -0,0 +1,343 @@
...
+#include "silabs_utils.h"
+#include "sl_constants.h"
+#include "sl_net.h"
+#include "sl_status.h"
+#include "stdbool.h"
+#include <stdlib.h>
+#include <string.h>
+
+#ifndef MQTT_USE_HOST_LWIP_TLS
+#define MQTT_USE_HOST_LWIP_TLS 0
```
agarwal57: - Can you please explain what this mean here? This is for the MQTTS vs MQTT? can you let me know what is the default here?

src/platform/silabs/services/mqtt_client.c
```diff
@@ -0,0 +1,343 @@
...
+
+#define MQTT_QOS 1
+#define TOPIC_TO_BE_SUBSCRIBED "THERMOSTAT-DATA"
+
+#define TCP_MQTT_CLIENT_INIT_BUFF_LEN 3500
+#define TCP_MQTT_CLIENT_RX_BUFFER_SIZE 1500
+#define TCP_MQTT_CLIENT_TX_BUFFER_SIZE 1500
+#define TCP_MQTT_CONNECT_TIME_OUT 20000
+#define KEEP_ALIVE_PERIOD 100
+#define MQTT_VERSION 4
```
agarwal57: - What is this MQTT VERSION? i think we are using 3.1.1. So, am wondering what this version means here.

src/platform/silabs/services/mqtt_client.c
```diff
@@ -0,0 +1,343 @@
...
+static osThreadId_t s_mqtt_demo_thread_id = NULL;
+
+static const osThreadAttr_t mqtt_lwip_thread_attributes = {
+    .name       = "mqtt_lwip",
+    .attr_bits  = 0,
+    .cb_mem     = 0,
+    .cb_size    = 0,
+    .stack_mem  = 0,
+    .stack_size = (8 * 1024), /* mbedTLS TLS1.2 + RSA-2048/PSA: measured ~1.2KB stack free before SKE verify (3KB stack). */
+    .priority   = osPriorityAboveNormal,
```
agarwal57: HTTPS is belowNormal and this is AboveNormal. Can you please help with the basis of defining this priority?

src/platform/silabs/services/mqtt_client.c
```diff
@@ -0,0 +1,343 @@
...
+    if (enable_ssl)
+    {
+        mqtt_client->client.ipstack->tls = malloc(sizeof(mqtt_tls_context_t));
+        if (!mqtt_client->client.ipstack->tls)
+        {
+            SILABS_LOG("MQTT error: failed to allocate TLS context");
+            return -1;
+        }
+        configure_tls_certificates(&mqtt_client->client.ipstack->tls->cert_ctx);
+        if (NetworkSetTlsHostname(mqtt_client->client.ipstack, MQTT_BROKER_TLS_HOSTNAME) != 0)
```
agarwal57: - There are comments missing in this function. Please add some comments separating sections in the entire logic of the functions, which are large. --> Generic comment and please handle in all places where appropriate.

src/platform/silabs/services/mqtt_client.c
```diff
@@ -0,0 +1,343 @@
...
+    status = MQTTConnect(&mqtt_client->client, &connectData);
+    if (status != 0)
+    {
+        SILABS_LOG("MQTT error: MQTT connect failed: %d", status);
+        NetworkDisconnect(mqtt_client->client.ipstack);
+        return status;
+    }
+    SILABS_LOG("MQTT connected");
+
+    status = MQTTSubscribe(&mqtt_client->client, (char *) TOPIC_TO_BE_SUBSCRIBED, (enum QoS) MQTT_QOS, message_arrived);
```
agarwal57: - rename the message_arrived to subscription_callback.

src/platform/silabs/services/mqtt_client.c
```diff
@@ -0,0 +1,343 @@
...
+        publish_msg.qos = QOS0;
+    }
+    else if (MQTT_QOS == QOS1)
+    {
+        publish_msg.qos = QOS1;
+    }
+    else
+    {
+        publish_msg.qos = QOS2;
+    }
```
agarwal57: - We are supporting all the QoS for MQTT and default is QOS2, if nothing is given?

src/platform/silabs/wifi/SiWx/WifiInterfaceImpl.cpp
```diff
@@ -36,6 +36,12 @@
 #include <platform/silabs/wifi/SiWx/WifiInterfaceImpl.h>
 #include <sl_cmsis_os2_common.h>
 
+#include <sl_mbedtls_config.h>
+
+#ifdef MBEDTLS_PSA_CRYPTO_C
+#include <psa/crypto.h>
+#endif // MBEDTLS_PSA_CRYPTO_C
```
agarwal57, line 43: - we require this? Do we have info on why is this required?

agarwal57 (changes requested):
Some of the comments given in the HTTPS are applicable here as well including the one making this service long lived instead of dying after doing a specific action.

src/lwip/silabs/lwipopts-siwx.h
rosahay-silabs: `SLI_PAHO_MQTT_OVER_LWIP` this we are picking up from WiseConnect, I do understand the concern. Let me see if I can come up with something here instead.

examples/platform/silabs/BaseApplication.cpp
rosahay-silabs: done

examples/platform/silabs/MatterShell.cpp
rosahay-silabs: added comment

examples/platform/silabs/freertos_config/FreeRTOSConfig.h
rosahay-silabs: the comment was added while some experimentation, removed it.

examples/platform/silabs/freertos_config/FreeRTOSConfig.h
rosahay-silabs: done

src/lwip/freertos/sys_arch.c
rosahay-silabs: done

src/platform/silabs/services/mqtt_client.c
rosahay-silabs: `MQTT_USE_HOST_LWIP_TLS` is this is not mentioned then the application is MQTT over LwIP, meanwhile if you have `MQTT_USE_HOST_LWIP_TLS` enabled then it is MQTTs over LwIP

src/platform/silabs/services/mqtt_client.c
rosahay-silabs: This is what was outlined on the example application by WiseConnect. We can change this based on the priority of the communication.

src/platform/silabs/services/mqtt_client.c
rosahay-silabs: `QoS1` by default

src/platform/silabs/wifi/SiWx/WifiInterfaceImpl.cpp
rosahay-silabs: Yes this is required for now, need to dig into it more to figure out if we can discard this.

examples/thermostat/silabs/src/mqtt_example.cpp
```diff
@@ -0,0 +1,233 @@
+/*
+ *
+ *    Copyright (c) 2026 Project CHIP Authors
```
jepenven-silabs, line 3: this files looks a lot like this one https://github.com/SiliconLabsSoftware/matter_sdk/pull/1146/changes#diff-ceb4aeda9a39235555af2d9abae44729b9944e44a426a08f6b79227d7f7d7b5a 

Wasn't there any way of minimizing code duplication with a base class/inheritance?

examples/platform/silabs/MatterShell.cpp
```diff
@@ -102,7 +102,12 @@ void cmdSilabsInit()
 
 #endif // SL_CATALOG_CLI_PRESENT
 
+#if defined(SLI_SI91X_MCU_INTERFACE) && SLI_SI91X_MCU_INTERFACE
```
jepenven-silabs: According to this PR we shouldn't need anymore `#if` for memory manager API...

https://github.com/SiliconLabsInternal/platform/commit/c0fd0aa06e208928c62fa70f44db1e8b6cfc02f3 

If this commit is not present on your working branch then I would simply `#ifdef` the memory shell commands instead of adding temporary `#if` accross the stack...

jepenven-silabs (changes requested):

examples/platform/silabs/MatterShell.cpp
rosahay-silabs: @jepenven-silabs, as far as I know this will come to 4.2.x and not 4.1.2 wiseconnect, this branch is based on 4.1.2 wiseconnect.

examples/platform/silabs/freertos_config/FreeRTOSConfig.h
```diff
@@ -282,6 +282,12 @@ See http://www.FreeRTOS.org/RTOS-Cortex-M3-M4.html. */
 #define EXTRA_HEAP_k 0
 #endif
 
+#ifdef MQTT_USE_HOST_LWIP_TLS
+#define MQTT_EXTRA_HEAP_k 28
```
agarwal57, line 286: - @rosahay-silabs, are we not requiring 60kB additionally for the MQTTS connection? Can we get this clarity and update accordingly? This will raise questions from customer where we mentioned that we require about 60kB for each SSL connection, and here we are putting only 28kB.

examples/thermostat/silabs/src/mqtt_example.cpp
```diff
@@ -0,0 +1,233 @@
...
+} // namespace
+
+extern "C" sl_status_t mqtt_client_demo_start(void)
+{
+    if (gMqttClient.IsRunning())
+    {
+        return SL_STATUS_ALREADY_INITIALIZED;
+    }
+
+    if (gDoneLock == nullptr)
```
agarwal57: - what is the purpose of semaphore for MQTT?

src/lwip/silabs/lwipopts-siwx.h
agarwal57: Any update on this?

src/platform/silabs/wifi/SiWx/WifiInterfaceImpl.cpp
agarwal57: Can we have a JIRA to investigate this to ensure that we are not missing... As suggested earlier, create one JIRA which we are deferring here and should be followed up with a next PR so that we do not miss it.

src/platform/silabs/services/mqtt_client.h
```diff
@@ -0,0 +1,243 @@
...
+#endif
+
+namespace chip {
+namespace DeviceLayer {
+namespace Silabs {
+
+/**
+ * @brief MQTT quality-of-service levels (maps to Paho QOS0 / QOS1 / QOS2).
+ */
+enum class MqttQos : uint8_t
```
agarwal57: - rename MqttQos to MqttQoS for better readability

src/platform/silabs/services/mqtt_client.h
```diff
@@ -0,0 +1,243 @@
...
+    const char * password           = nullptr;
+    uint16_t keepAliveIntervalSec   = 100;
+    unsigned int commandTimeoutMs   = 20000;
+    uint8_t mqttVersion             = 4;
+    bool cleanSession               = true;
+
+    bool willEnable                 = false;
+    const char * willTopic          = nullptr;
+    const char * willMessage        = nullptr;
+    MqttQos willQos                 = MqttQos::AtLeastOnce;
```
agarwal57: - lets keep QoS instead of Qos wherever applicable

src/platform/silabs/services/mqtt_client.h
```diff
@@ -0,0 +1,243 @@
...
+
+/**
+ * @brief Broker destination for @ref MqttClient::Connect.
+ */
+struct MqttBroker
+{
+    const char * brokerIp     = nullptr; ///< IPv4 string for sl_net_inet_addr. Must not be nullptr.
+    const char * tlsHostname  = nullptr; ///< SNI / cert verify name when TLS is enabled.
+    uint16_t brokerPort       = 0;       ///< Broker port (1883 or 8883 typical).
+    uint16_t clientPort       = 0;       ///< Local source port.
```
agarwal57: - do we need the client port to be tracked from the application side? What is this clientPort used for?

src/platform/silabs/services/mqtt_client.h
```diff
@@ -0,0 +1,243 @@
...
+ */
+using MqttOperationCallback = void (*)(CHIP_ERROR result, void * context);
+
+/**
+ * @brief Incoming publish handler.
+ *
+ * Invoked on the MqttClient service thread from Yield / Subscribe delivery.
+ * @p topic and @p payload are valid only for the duration of the call.
+ */
+using MqttMessageCallback = void (*)(const char * topic, ByteSpan payload, void * context);
```
agarwal57: - what is the standard? I believe MqttSubscriptionCallback would be more appropriate. Can you please check?

src/platform/silabs/services/mqtt_client.h
```diff
@@ -0,0 +1,243 @@
...
+
+    CHIP_ERROR Start() override;
+    CHIP_ERROR Stop() override;
+    bool IsRunning() const override;
+    bool IsBusy() const;
+
+    /**
+     * @brief Set handler for subscribed publishes. Safe to call before Connect.
+     */
+    void SetMessageCallback(MqttMessageCallback callback, void * context = nullptr);
```
agarwal57: - so, a single callback is expected for all the topics that are being subscribed? Is this to keep it simple or this design is recommended? So, in case if the user wants to have a different callback for different topics? What would be the required changes?

src/platform/silabs/services/mqtt_client.h
```diff
@@ -0,0 +1,243 @@
...
+
+    /**
+     * @brief Run MQTTYield for up to @p timeoutMs (delivers messages via @ref SetMessageCallback).
+     */
+    CHIP_ERROR Yield(uint32_t timeoutMs, MqttOperationCallback callback, void * context = nullptr);
+
+private:
+    static constexpr size_t kTxBufferSize           = 1500;
+    static constexpr size_t kRxBufferSize           = 1500;
+    static constexpr size_t kDefaultThreadStackSize = 8 * 1024;
```
agarwal57, line 171: - here as well, how did we arrive at the 8kB stack size? Any data we have for this. Can you please add comments

src/platform/silabs/services/mqtt_client.cpp
```diff
@@ -0,0 +1,611 @@
...
+    return mBusy;
+}
+
+void MqttClient::SetMessageCallback(MqttMessageCallback callback, void * context)
+{
+    mMessageCallback        = callback;
+    mMessageCallbackContext = context;
+}
+
+void MqttClient::PahoMessageHandler(MessageData * md)
```
agarwal57, line 74: - No comments on for this function.

Again generic comment to ensure that add comments to the functions briefly describing the purpose of the functions, params and return values. Also, add comments with in the functions as we discussed.... If a function does 3 to 4 things at a high-level, at least those 3 to 4 comments for each of those steps/things is necessary for readability and future maintainability.

src/platform/silabs/services/mqtt_client.cpp
```diff
@@ -0,0 +1,611 @@
...
+{
```
agarwal57, line 75: - looks like this is more of parsing the MQTT message and passing the contents like topic and payload to the callback.... The comments will help the reviewer to undersatnd what is coming next... Here we need to guess what the function does after gonig through the function as sometimes names are not very intuitive.

src/platform/silabs/services/mqtt_client.cpp
```diff
@@ -0,0 +1,611 @@
...
+
+    mPendingOperation    = operation;
+    mUserCallback        = callback;
+    mUserCallbackContext = context;
+    mBusy                = true;
+    osEventFlagsSet(static_cast<osEventFlagsId_t>(mEventFlags), kEventOperation);
+    return CHIP_NO_ERROR;
+}
+
+void MqttClient::CompleteOperation(CHIP_ERROR error)
```
agarwal57, line 225: - this looks to be overcomplicating things. What is required here? Actually why do you need to even have topics, callbacks etc variables as they will be directly coming from the application right? 

- Also, are the calls in PahoMQTT are async and timetaking?

agarwal57 (changes requested):
Overall, i find this complicated and my opinion is that we need to revisit and simplify stuff. You can consider some of the feedback from HTTPS as well here as applicable. But as we are nearing our timeline, lets do the necessary changes and identify things to revisit.

src/platform/silabs/services/mqtt_client.h
sabollim-silabs: we can adopt QoS0, 1,2 model right why this naming convention

src/platform/silabs/services/mqtt_client.h
```diff
@@ -0,0 +1,243 @@
...
+    const uint8_t * tlsCaCert = nullptr;
+    size_t tlsCaCertLen       = 0;
+};
+
+/**
+ * @brief Broker destination for @ref MqttClient::Connect.
+ */
+struct MqttBroker
+{
+    const char * brokerIp     = nullptr; ///< IPv4 string for sl_net_inet_addr. Must not be nullptr.
```
sabollim-silabs: its good  to take broker url also and fetch ip using DNS

src/platform/silabs/services/mqtt_client.h
```diff
@@ -0,0 +1,243 @@
...
+    bool useTls                     = true;
+    const char * clientId           = nullptr;
+    const char * username           = nullptr;
+    const char * password           = nullptr;
+    uint16_t keepAliveIntervalSec   = 100;
+    unsigned int commandTimeoutMs   = 20000;
+    uint8_t mqttVersion             = 4;
+    bool cleanSession               = true;
+
+    bool willEnable                 = false;
```
sabollim-silabs: what does this used for and any speacific reason of will* in naming?

examples/platform/silabs/freertos_config/FreeRTOSConfig.h
rosahay-silabs: Hey @agarwal57, based on the calculations so far, we have not seen a requirement for 60 kB of heap. We are currently allocating 100 kB of total heap, and with both examples running concurrently, the peak heap usage is around 97 kB. 60 kB heap requirement was not something we committed to from the Matter side, it was an estimate provided by the WiseConnect team. I can increase the allocation to 60 kB if we suspect that the current value could cause issues during stability testing.
That said, the testing with these changes has been limited so far. We would need to run the full test suite with both clients active in the background before we can confidently validate whether 28 kB is sufficient.

src/lwip/silabs/lwipopts-siwx.h
rosahay-silabs: Though these settings are specific to the PAHO MQTT that is bundled in WiseConnect, I am adding `SL_MATTER_ENABLE_MQTT_OVER_LWIP_SOCKETS` for better clarity.

src/platform/silabs/services/mqtt_client.h
rosahay-silabs: done

rosahay-silabs: done

src/platform/silabs/services/mqtt_client.h
rosahay-silabs: done

src/platform/silabs/services/mqtt_client.h
rosahay-silabs: Agreed, but that is planned for the next delivery.

src/platform/silabs/services/mqtt_client.h
rosahay-silabs: Typically should be configurable, it is used to `bind` client socket for the subscription callback. It can be ephemeral in nature, unlike the `brokerPort`, but since examples configure it to `5003` I did the same.

src/platform/silabs/services/mqtt_client.h
rosahay-silabs: Going by the PAHO MQTT bundled, it is called `messageHandler` and not subscription handler, but we can rename it to fit our needs.

src/platform/silabs/services/mqtt_client.h
rosahay-silabs: To simplify the implementation a single callback was done, this needs to be improved with a default message handler, and per subscription message handler.
Let's track it as part of feature complete.

src/platform/silabs/wifi/SiWx/WifiInterfaceImpl.cpp
rosahay-silabs: This is required as a component is getting added from the `sl_si91x_psa_trng` → `psa_crypto`, sure lets add this to the JIRA to investigate what can be done to remove this dependency.

examples/thermostat/silabs/src/mqtt_example.cpp
rosahay-silabs: different examples, which can exist independently, apart from the hostname port and enableTLS options, there is not much common. On a very high level yes, we can make a base example for `connect` `disconnect` and a generic `callback` but I do not see that much overlap, can you help me understand what is suggested? But since it is a sample example I did not put that much thought into it.

src/platform/silabs/services/mqtt_client.h
rosahay-silabs: This was mostly reached via experimentation on the THREAD stack size with wiseconnect demo and integration.

src/platform/silabs/services/mqtt_client.cpp
rosahay-silabs: >this looks to be overcomplicating things. What is required here? Actually why do you need to even have topics, callbacks etc variables as they will be directly coming from the application right?

This client was not meant to be a transparent proxy.

Currently, my understanding goes, that the PAHO MQTT client is running on this THREAD context. Right now, if the consumer, invokes a MQTTs API then it needs to be put into the message queue(operation queue) with the other details that the consumer has passed, that needs to be available once the operation is to be processed. I will verify this once with the code flow, currently documentation is kind of scarce around this.

>Also, are the calls in PahoMQTT are async and timetaking?

Publish looks to be time taking in the experimentation, but further analysis is required.

src/platform/silabs/services/mqtt_client.h
rosahay-silabs: inherited the naming from the structure on the WiseConnect, but will take time to investigate this and revert.

src/platform/silabs/services/mqtt_client.h
```diff
@@ -0,0 +1,243 @@
...
+#include <lib/support/Span.h>
+
+#include <cstddef>
+#include <cstdint>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+#include "MQTTClient.h"
+#include "sl_net.h"
```
jepenven-silabs, line 35: sl_net.h and sl_status.h shouldn't need to be inside this `extern "C"` If this is the case, a bug should be open against the team that owns the file...

src/platform/silabs/services/mqtt_client.cpp
```diff
@@ -0,0 +1,611 @@
...
+}
+
+void MqttClient::PahoMessageHandler(MessageData * md)
+{
+    MqttClient * self = sActiveClient;
+    VerifyOrReturn(self != nullptr);
+    VerifyOrReturn(md != nullptr && md->message != nullptr);
+
+    const char * topic = nullptr;
+    char topicBuf[128] = { 0 };
```
jepenven-silabs: magic number

sabollim-silabs (approved):
few comments are not yet addressed, approving the PR as they can be addressed in follow up PRs

src/platform/silabs/services/mqtt_client.cpp
```diff
@@ -0,0 +1,611 @@
...
+    else
+    {
+        ChipLogProgress(DeviceLayer, "MQTT message: %.*s", static_cast<int>(payload.size()),
+                        reinterpret_cast<const char *>(payload.data()));
+    }
+}
+
+void MqttClient::LogNetworkConnectError(int status)
+{
+    if (status == NETWORK_ERROR_NULL_STRUCTURE)
```
jepenven-silabs: NIT : A switch case function with a single log would have been easier to read. 
you can even append the error code at the end of the string if needed


```suggestion
    char * getErrorString(int status)
    {
        switch(status) {
            case NETWORK_ERROR_NULL_STRUCTURE : 
                return "MQTT network connect failed on dual-stack use paho_mqtt_embedded with dual-stack network"
             case NETWORK_ERROR_NULL_ADDRESS:
                  return "MQTT broker address is NULL"
        }
    }
    void MqttClient::LogNetworkConnectError(int status)
    {
        ChipLogError(DeviceLayer, "%s", getErrorString(status));
    }
    
```

src/platform/silabs/services/mqtt_client.cpp
```diff
@@ -0,0 +1,611 @@
...
+void MqttClient::ServiceThread(void * arg)
+{
+    auto * self = static_cast<MqttClient *>(arg);
+    VerifyOrReturn(self != nullptr);
+
+    auto eventFlags = static_cast<osEventFlagsId_t>(self->mEventFlags);
+    while (true)
+    {
+        uint32_t events = osEventFlagsWait(eventFlags, kEventOperation | kEventStop, osFlagsWaitAny, osWaitForever);
+        if ((events & osFlagsError) != 0)
```
jepenven-silabs: So all errors are treated as critical failure ?

src/platform/silabs/services/mqtt_client.cpp
```diff
@@ -0,0 +1,611 @@
...
+        {
+            break;
+        }
+        if ((events & kEventOperation) != 0)
+        {
+            self->ProcessOperation();
+        }
+    }
+
+    self->mThreadId = nullptr;
```
jepenven-silabs: Within `mqtt_client.h` there is a lot of pointers..... and the cleanup in case of failure seems pretty small? are we sure that everything that needs to be freed is in indeed freed properly ? In mean what happens to the osFlag in case of a restart  of the thread ?

src/platform/silabs/services/mqtt_client.cpp
```diff
@@ -0,0 +1,611 @@
...
+
+    VerifyOrReturnError(!mBusy, CHIP_ERROR_BUSY);
+
+    osEventFlagsSet(static_cast<osEventFlagsId_t>(mEventFlags), kEventStop);
+    while (mThreadId != nullptr)
+    {
+        osDelay(1);
+    }
+
+    osEventFlagsDelete(static_cast<osEventFlagsId_t>(mEventFlags));
```
jepenven-silabs: Inside which context is the call to stop made ? should it be thread itself that deletes it's own resources ?

src/platform/silabs/services/mqtt_client.cpp
```diff
@@ -0,0 +1,611 @@
...
+CHIP_ERROR MqttClient::QueueOperation(Operation operation, MqttOperationCallback callback, void * context)
+{
+    VerifyOrReturnError(IsRunning(), CHIP_ERROR_INCORRECT_STATE);
+    VerifyOrReturnError(!mBusy, CHIP_ERROR_BUSY);
+
+    mPendingOperation    = operation;
+    mUserCallback        = callback;
+    mUserCallbackContext = context;
+    mBusy                = true;
+    osEventFlagsSet(static_cast<osEventFlagsId_t>(mEventFlags), kEventOperation);
```
jepenven-silabs, line 221: Considering that the thread can be stopped and resources can be deleted a check for `nullptr` or at least for the return of each OS API should be needed.

src/platform/silabs/services/mqtt_client.cpp
```diff
@@ -0,0 +1,611 @@
...
+    VerifyOrReturnError(mConfig.clientId != nullptr, CHIP_ERROR_INVALID_ARGUMENT);
+
+    memset(&mClient, 0, sizeof(mClient));
+    memset(&mNetwork, 0, sizeof(mNetwork));
+    memset(mTxBuffer, 0, sizeof(mTxBuffer));
+    memset(mRxBuffer, 0, sizeof(mRxBuffer));
+    memset(&mServerIp, 0, sizeof(mServerIp));
+
+    mNetwork.transport_type = MQTT_TRANSPORT_TCP;
+    NetworkInit(&mNetwork);
```
jepenven-silabs, line 294: I find it strange that both the `NetworkInit` and that the `MQTTClient` cannot fail.... If they can, then we should validate the return code.

jepenven-silabs (approved):
Approve pending comment resolution/answer. Resolution can be in a follow up PR.

src/platform/silabs/services/mqtt_client.cpp
rosahay-silabs: resolved

src/platform/silabs/services/mqtt_client.cpp
rosahay-silabs: done

src/platform/silabs/services/mqtt_client.cpp
rosahay-silabs: yes for now, we will have to refine it, for all services, maybe make it similar to how matter handles it.

src/platform/silabs/services/mqtt_client.cpp
rosahay-silabs: it was added in the `Stop` function, but have moved it here instead, the `kEventStop` will terminate the event loop, and the clean up of event flags and thread will be done.

src/platform/silabs/services/mqtt_client.cpp
rosahay-silabs: yes, moved the context to the THREAD itself now.

src/platform/silabs/services/mqtt_client.cpp
rosahay-silabs: done

src/platform/silabs/services/mqtt_client.cpp
rosahay-silabs: Since I also did not deep dive into the codebase, it looks like both `NetworkInit` and `MQTTClient` are `void` functions, so I am under the impression that anything goes wrong here only shows up while connecting to network in the `NetworkConnect`

src/platform/silabs/services/mqtt_client.h
rosahay-silabs: done

PR #1146, opened 2026-08-13
[SL-ONLY] MATTER-6537: Adds NWP HTTPs service demo

rosahay-silabs:
### Summary

- Add NWP offload HTTPS client demo (`http_client.c` / `http_client.h`) with CA cert embed and PUT/GET/POST against a local HTTPS server
- Start `http_client_demo` from `BaseApplication` after IPv4 (or dual-stack IPv6) connectivity when Matter services / HTTP service are enabled
- Enable the SiWx `SL_SI91X_TCP_IP_FEAT_HTTP_CLIENT` feature bit when the HTTP service is selected

### Related issues

MATTER-6537

### Testing

- Tested the WiseConnect example after integration with MATTER using `simple-https-server.py`
- Executed basic MATTER commands


src/platform/silabs/services/cacert.pem.h
```diff
@@ -0,0 +1,41 @@
+/*******************************************************************************
+ * @file  cacert.pem.h
+ * @brief Auto-generated by generate_mqtt_certs.py
+ *******************************************************************************
+ * # License
+ * <b>Copyright 2023 Silicon Laboratories Inc. www.silabs.com</b>
```
sabollim-silabs:
```suggestion
 * <b>Copyright 2026 Silicon Laboratories Inc. www.silabs.com</b>
```

src/platform/silabs/services/http_client.c
```diff
@@ -0,0 +1,464 @@
...
+
+#define HTTP_CLIENT_USERNAME "admin"
+#define HTTP_CLIENT_PASSWORD "admin"
+
+#define HTTP_IP_VERSION SL_IPV4
+#define HTTP_VERSION SL_HTTP_V_1_1
+#define HTTP_TLS_VERSION SL_TLS_V_1_2
+#define HTTP_CERT_INDEX SL_HTTPS_CLIENT_CERTIFICATE_INDEX_1
+
+#define HTTP_SERVER_IP "192.168.0.191"
```
sabollim-silabs: should these be hardcoded

sabollim-silabs: same comment or username and password

sabollim-silabs (approved):
LGTM, did go into the functionality of https_client.c as the code is leveraged from wifi app

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -1053,6 +1071,18 @@ void BaseApplication::OnPlatformEvent(const ChipDeviceEvent * event, intptr_t)
                 chip::System::Clock::Seconds32(MATTER_AWS_INIT_DELAY_SEC), InitMatterAwsHandler, nullptr);
         }
 #endif // SL_MATTER_ENABLE_AWS
+#ifdef SL_MATTER_ENABLE_SERVICES
```
agarwal57: This entire block looks to be duplicate of the above one for SL_MATTER_ENABLE_AWS macro. Can you please combine this into one by checking either of the macros may be by replace ifdef with if defined(macro1) || defined(macro2)

src/platform/silabs/services/http_client.c
```diff
@@ -0,0 +1,464 @@
+/***************************************************************************/
+/**
+ * @file http_client.c
+ * @brief NWP offload HTTPS client demo (invoked from Matter)
+ *******************************************************************************
+ * # License
+ * <b>Copyright 2025 Silicon Laboratories Inc. www.silabs.com</b>
```
agarwal57: - update the year here as well.

src/platform/silabs/services/http_client.c
```diff
@@ -0,0 +1,464 @@
...
+#include <string.h>
+
+#define TLS_CERTIFICATE_INDEX 1
+
+#define HTTP_CLIENT_USERNAME "admin"
+#define HTTP_CLIENT_PASSWORD "admin"
+
+#define HTTP_IP_VERSION SL_IPV4
+#define HTTP_VERSION SL_HTTP_V_1_1
+#define HTTP_TLS_VERSION SL_TLS_V_1_2
```
agarwal57: - This is supported for TLS 1.3 as well right? How can this be configured? Should we by default go with the highest version? What is the recommendation from Wiseconnect on this?

src/platform/silabs/services/http_client.c
agarwal57: @rosahay-silabs, these are not addressed... Please check

src/platform/silabs/services/http_client.c
```diff
@@ -0,0 +1,464 @@
...
+#define HTTP_SERVER_IP "192.168.0.191"
+#define HTTP_HOSTNAME "192.168.0.191"
+#define HTTP_PORT 8443
+
+#define HTTP_URL "/index.html"
+#define HTTP_DATA                                                                                                                  \
+    "employee_name=MR.REDDY&employee_id=RSXYZ123&designation=Engineer&company="                                                    \
+    "SILABS&location=Hyderabad"
+
+#define APP_BUFFER_LENGTH 2000
```
agarwal57: - Can you add a comment here on what this APP_BUFFER_LENGTH mean here?

src/platform/silabs/services/http_client.c
```diff
@@ -0,0 +1,464 @@
...
+
+#define HTTP_SUCCESS_RESPONSE 1
+#define HTTP_FAILURE_RESPONSE 2
```
agarwal57: - Are these standard responses of HTTP? If not, shouldnt this be kinda 0 and 1 the standard ones which we use for success and failure?

src/platform/silabs/services/http_client.c
```diff
@@ -0,0 +1,464 @@
...
+
+#define HTTP_STATUS_CLIENT_ERROR_MIN 400U
+#define HTTP_STATUS_SERVER_ERROR_MAX 599U
+#define HTTP_STATUS_CODE_NONE 0U
```
agarwal57: What is the HTTP status code none mean? And the purpose of this?

src/platform/silabs/services/http_client.c
```diff
@@ -0,0 +1,464 @@
...
+static osThreadId_t s_https_demo_thread_id = NULL;
+static bool s_nwp_ca_loaded                = false;
+
+static const osThreadAttr_t https_offload_thread_attributes = {
+    .name       = "https_offload",
+    .attr_bits  = 0,
+    .cb_mem     = 0,
+    .cb_size    = 0,
+    .stack_mem  = 0,
+    .stack_size = (2 * 1024),
```
agarwal57: - how is this stack size determined?

src/platform/silabs/services/http_client.c
```diff
@@ -0,0 +1,464 @@
...
+    }
+
+    s_nwp_ca_loaded = true;
+    SILABS_LOG("HTTPS loaded TLS CA at index %d", TLS_CERTIFICATE_INDEX);
+    return SL_STATUS_OK;
+}
+
+static sl_status_t http_response_status(volatile uint8_t * response)
+{
+    while (*response != HTTP_SUCCESS_RESPONSE && *response != HTTP_FAILURE_RESPONSE)
```
agarwal57: Can you please add comments to the code? Generic comment to add comments to all functions added newly as part of this PR.

agarwal57: In particular in this case, am not clear on what other values are possible and when do we put them as 1 and 2 here

src/platform/silabs/services/http_client.c
```diff
@@ -0,0 +1,464 @@
...
+
+    status = sl_http_client_init(&client_configuration, &client_handle);
+    if (status != SL_STATUS_OK)
+    {
+        free(client_credentials);
+        return status;
+    }
+    SILABS_LOG("HTTPS client init success");
+
+    sl_http_client_tcp_tls_advanced_options_t tcp_tls_opts = {
```
agarwal57: - All these are hardcoded values need to be defined as the constants.
- am assuming these will be part of the configuration for the http client we will instantiate.

agarwal57 (changes requested):

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -89,6 +89,14 @@
 #include "MatterAwsControl.h"
 #endif // SL_MATTER_ENABLE_AWS
 
+#ifdef SL_MATTER_ENABLE_SERVICES
```
sabollim-silabs: #ifdef works for zero value also right , can we guard it similar to line 94 if defined(macro) && macro

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -1029,6 +1037,16 @@ void InitMatterAwsHandler(System::Layer * systemLayer, void * appState)
 } // namespace
 #endif // SL_MATTER_ENABLE_AWS
 
+#ifdef SL_MATTER_ENABLE_SERVICES
```
sabollim-silabs: #ifdef works for zero value also right , can we guard it similar to line 94 if defined(macro) && macro

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -1029,6 +1037,16 @@ void InitMatterAwsHandler(System::Layer * systemLayer, void * appState)
...
+namespace {
+void InitMatterServicesHandler(System::Layer * systemLayer, void * appState)
+{
+#ifdef SL_MATTER_ENABLE_HTTP_SERVICE
```
sabollim-silabs: #ifdef works for zero value also right , can we guard it similar to line 94 if defined(macro) && macro

sabollim-silabs: same comment applies for all #ifdef

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -1053,6 +1071,18 @@ void BaseApplication::OnPlatformEvent(const ChipDeviceEvent * event, intptr_t)
...
 #endif // SL_MATTER_ENABLE_AWS
+#ifdef SL_MATTER_ENABLE_SERVICES
+        if (event->InternetConnectivityChange.IPv4 == kConnectivity_Established
+#if defined(SL_MATTER_ENABLE_DUAL_STACK) && SL_MATTER_ENABLE_DUAL_STACK
+            || event->InternetConnectivityChange.IPv6 == kConnectivity_Established
+#endif
+        )
+        {
+            ChipLogProgress(AppServer, "Scheduling Matter Services initialization");
+            RETURN_SAFELY_IGNORED chip::DeviceLayer::SystemLayer().StartTimer(chip::System::Clock::Seconds32(10),
```
sabollim-silabs: lets have a macro for this delay value

sabollim-silabs:
bugbot review

examples/platform/silabs/BaseApplication.cpp
rosahay-silabs: I do not plan to keep it here since as per conversion with @jmartinez-silabs this should be moved into example code where we expect that `init` to be handled by the customer, but since wiring the signals for all events are pending, this is kept here.

src/platform/silabs/services/http_client.c
rosahay-silabs: This has been addressed by moving these values into the demo code instead of the client, in #1149

examples/platform/silabs/BaseApplication.cpp
rosahay-silabs: This is temporary code here, and this code will move into the responsibility of the customer after wiring of the events.
`SL_MATTER_ENABLE_AWS` is something we will deprecate, since `SL_MATTER_ENABLE_<SERVICE_NAME>` should be used going forward. AWS will be an endpoint that we will test and support against, officially right?

src/platform/silabs/services/http_client.c
rosahay-silabs: I will check with WiseConnect on this and revert, but considering the example `SL_TLS_V_1_2` was carried over.

src/platform/silabs/services/http_client.c
rosahay-silabs: it is mostly calculated right now based on intuition and trial and error.

src/platform/silabs/services/http_client.c
rosahay-silabs: `app_buffer` with the `APP_BUFFER_LENGTH` is used as a buffer for the GET and PUT requests for the demo code, this has been modified in #1149

src/platform/silabs/services/http_client.c
rosahay-silabs: I have carried over the logic from the demo code to understand generic failure/success response, but I guess we also have the HTTP standard error codes to verify this, let me check the possibility of it.

src/platform/silabs/services/http_client.c
rosahay-silabs: `HTTP_STATUS_CODE_NONE` removed in the later PR #1149

src/platform/silabs/services/http_client.c
rosahay-silabs: Yes, as part of #1149 they have been taken into consideration but will review if we can move more of them out

src/platform/silabs/services/http_client.c
rosahay-silabs: As discussed we will test `SL_TLS_V_1_3` after verification with WiseConnect SQA.

src/platform/silabs/services/http_client.c
rosahay-silabs: This looks to be a limitation, analysing further with wiseconnect on this.

src/platform/silabs/services/http_client.c
rosahay-silabs: As mentioned this looks to be a limitation with `PUT` operation for the SiWx917 family of chips, will investigate further. Currently implemented as per the reference of the demo application. Attaching the snippet here

```cpp
typedef struct {
  uint32_t status;      ///< Status of the HTTP request.
  uint8_t *data_buffer; ///< Pointer to the buffer containing the response data.
  uint16_t data_length; ///< Length of the received data in bytes.
  uint32_t end_of_data; ///< Indicates the completion status of a data chunk or communication phase.
    ///<
    ///< GET/POST (indications from server):
    ///<   0: More response data is expected.
    ///<   1: This is the final chunk of the response.
    ///<
    ///< PUT:
    ///<   Acknowledgment for transmitted data (notification from NWP):
    ///<     0: Further data transmission is expected.
    ///<     1: Data transmission is complete.
    ///<   Server response (data received from server after transmission):
    ///<     8: More response data is expected.
    ///<     9: This is the final chunk of the response. See @ref SL_HTTP_CLIENT_PUT_SERVER_RESPONSE_END_OF_DATA.
  uint16_t
    http_response_code; ///< HTTP response code from the server. (Si91x chipsets do not support this feature for SL_HTTP_PUT).
  uint8_t version; ///< HTTP version used in the response. (Si91x chipsets do not support this feature).
  sl_http_client_header_t *
    response_headers; ///< Pointer to the HTTP response headers. See @ref sl_http_client_header_t. (Si91x chipsets do not support this feature).
} sl_http_client_response_t;
```

src/platform/silabs/services/http_client.c
rosahay-silabs: Currently going with defaults, will have to figure out if all clients support such options or not, based on additional inputs will decide if this will be another API to configure just for the SiWx917 or can be induced into the `HttpClientConfig`

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -1053,6 +1071,18 @@ void BaseApplication::OnPlatformEvent(const ChipDeviceEvent * event, intptr_t)
...
 #endif // SL_MATTER_ENABLE_AWS
+#if defined(SL_MATTER_ENABLE_SERVICES) && SL_MATTER_ENABLE_SERVICES
+        if (event->InternetConnectivityChange.IPv4 == kConnectivity_Established
+#if defined(SL_MATTER_ENABLE_DUAL_STACK) && SL_MATTER_ENABLE_DUAL_STACK
+            || event->InternetConnectivityChange.IPv6 == kConnectivity_Established
+#endif
+        )
+        {
+            ChipLogProgress(AppServer, "Scheduling Matter Services initialization");
+            RETURN_SAFELY_IGNORED chip::DeviceLayer::SystemLayer().StartTimer(chip::System::Clock::Seconds32(10),
```
jepenven-silabs: If we failed to init the services then something is definitively wrong. Would it be cause to simply assert ?

examples/thermostat/silabs/certs/cacert.pem.h
```diff
@@ -0,0 +1,38 @@
...
+ * The licensor of this software is Silicon Laboratories Inc. Your use of this
+ * software is governed by the terms of Silicon Labs Master Software License
+ * Agreement (MSLA) available at
+ * www.silabs.com/about-us/legal/master-software-license-agreement. This
+ * software is distributed to you in Source Code format and is governed by the
+ * sections of the MSLA applicable to Source Code.
+ *
+ ******************************************************************************/
+
+const unsigned char cacert[] = "-----BEGIN CERTIFICATE-----\r\n"
```
jepenven-silabs: I'm not a big fan of plain text cert in general, demo or not. But if we need to keep this one, then we should at least make the variable explicit on the fact the it contains an example cert. maybe with renaming it to `example_cacert` or `demo_cacert` ?

examples/thermostat/silabs/include/https_offload_example.h
```diff
@@ -0,0 +1,21 @@
...
+ ******************************************************************************/
+#pragma once
+
+#include "sl_status.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/** Load config + start HTTPS offload demo on a dedicated thread (idempotent). */
```
jepenven-silabs: comment says `HTTPS` but the function name is `HTTP_` ... they should match IMO

examples/thermostat/silabs/src/https_offload_example.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+namespace {
+
+using chip::ByteSpan;
+using chip::MutableByteSpan;
+using chip::DeviceLayer::Silabs::HttpClient;
+using chip::DeviceLayer::Silabs::HttpClientConfig;
+using chip::DeviceLayer::Silabs::HttpHost;
+
+constexpr char kHttpServerIp[]   = "192.168.0.191";
```
jepenven-silabs: Hardcoding the IPv4 address of the server here seems like a bad way of making this example work on any network. This deserved at least a comment like 

````
// change this address with the server address on your local network

File: examples/thermostat/silabs/src/https_offload_example.cpp [hunk 20]
```diff
@@ -0,0 +1,200 @@
...
+{
+    auto * waitCtx  = static_cast<OperationWaitContext *>(context);
+    waitCtx->result = result;
+    osSemaphoreRelease(waitCtx->lock);
+}
+
+CHIP_ERROR WaitForCallback(OperationWaitContext & waitCtx)
+{
+    const osStatus_t status = osSemaphoreAcquire(waitCtx.lock, osWaitForever);
+    VerifyOrReturnError(status == osOK, CHIP_ERROR_INTERNAL);
```
jepenven-silabs, line 89: This check is useless since we wait forever. Unless this function is called before the scheduler is running or when the scheduler is stopped it can/should be removed.

examples/thermostat/silabs/src/https_offload_example.cpp
```diff
@@ -0,0 +1,200 @@
...
+    }
+    ChipLogProgress(DeviceLayer, "HTTPS POST request success");
+
+    ChipLogProgress(DeviceLayer, "HTTPS demo completed");
+    return CHIP_NO_ERROR;
+}
+
+} // namespace
+
+extern "C" sl_status_t http_client_demo_start(void)
```
jepenven-silabs: why does this function needs to be an `extern "C"` Does it override a wiseconnect implementation ? if so a comment would be nice.

jepenven-silabs (changes requested):
Some minor comments left and right, but the biggest one that needs fixing IMO is the following. 

HTTP and HTTPS are two different things since one needs encryption and the other don't. This PR description, the comments in the code, the file names and the function prototypes seems to mix and match both HTTP and HTTPS without any concern as if they were the same protocol. This needs some sort of cleanup...

src/platform/silabs/services/http_client.h
```diff
@@ -0,0 +1,300 @@
...
+     * @retval CHIP_NO_ERROR              Request queued.
+     * @retval CHIP_ERROR_BUSY            Another request is in progress.
+     * @retval CHIP_ERROR_INCORRECT_STATE Client not initialized.
+     * @retval CHIP_ERROR_INVALID_ARGUMENT Null host or resource.
+     */
+    CHIP_ERROR Put(const HttpHost & host, const char * resource, ByteSpan body, HttpOperationCallback callback,
+                   void * context = nullptr);
+
+private:
+    static constexpr size_t kDefaultThreadStackSize = 4 * 1024;
```
agarwal57, line 208: - How did we arrive at this stack size of 4kB? i mean it was 2kB earlier and now updated to 4kB?

examples/thermostat/silabs/certs/cacert.pem.h
rcasallas-silabs: Since this constant is only used by `https_offload_example.cpp`, I think that's where it should be defined, in an anonymous namespace, and with a descriptive name, like `kHttpClientTlsCert`.

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -1029,6 +1037,16 @@ void InitMatterAwsHandler(System::Layer * systemLayer, void * appState)
 } // namespace
 #endif // SL_MATTER_ENABLE_AWS
 
+#if defined(SL_MATTER_ENABLE_SERVICES) && SL_MATTER_ENABLE_SERVICES
+namespace {
+void InitMatterServicesHandler(System::Layer * systemLayer, void * appState)
+{
+#if defined(SL_MATTER_ENABLE_HTTP_SERVICE) && SL_MATTER_ENABLE_HTTP_SERVICE
+    VerifyOrReturn(SL_STATUS_OK == http_client_demo_start(), ChipLogError(AppServer, "http_client_demo_start failed"));
```
rcasallas-silabs: This is a common base for all examples, but `http_client_demo_start` is both declared and defined in `thermostat/silabs`. Other examples cannot be compiled with SL_MATTER_ENABLE_HTTP_SERVICE. Either `http_client_demo_start` must be called from the thermostat init, or the function should be part of the common code.

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -1029,6 +1037,16 @@ void InitMatterAwsHandler(System::Layer * systemLayer, void * appState)
 } // namespace
 #endif // SL_MATTER_ENABLE_AWS
 
+#if defined(SL_MATTER_ENABLE_SERVICES) && SL_MATTER_ENABLE_SERVICES
```
rcasallas-silabs, line 1042: `SL_MATTER_ENABLE_SERVICES` and `SL_MATTER_ENABLE_HTTP_SERVICE` doesn't seem to be defined anywhere. These should be defined as 0 or false, with a proper doc header, in a configuration file.

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -1029,6 +1037,16 @@ void InitMatterAwsHandler(System::Layer * systemLayer, void * appState)
...
+namespace {
+void InitMatterServicesHandler(System::Layer * systemLayer, void * appState)
+{
+#if defined(SL_MATTER_ENABLE_HTTP_SERVICE) && SL_MATTER_ENABLE_HTTP_SERVICE
+    VerifyOrReturn(SL_STATUS_OK == http_client_demo_start(), ChipLogError(AppServer, "http_client_demo_start failed"));
+#endif // SL_MATTER_ENABLE_HTTP_SERVICE
+}
+} // namespace
+#endif // SL_MATTER_ENABLE_SERVICES
 void BaseApplication::OnPlatformEvent(const ChipDeviceEvent * event, intptr_t)
```
rcasallas-silabs, line 1053: Please leave some space before, and after each function declaration. I find these dense blocks of code hard to read

examples/platform/silabs/BaseApplication.cpp
jmartinez-silabs: similar thought here. We should always define our defines and then use
`#if XYZ `or `#if XYZ == 0` depending on the situation.
I know that we can't do that with some defines that come from our sdks and they aren't always define. But the ones we add and have control should follow this standard. It is in our coding standard

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -89,6 +89,14 @@
 #include "MatterAwsControl.h"
 #endif // SL_MATTER_ENABLE_AWS
 
+#if defined(SL_MATTER_ENABLE_SERVICES) && SL_MATTER_ENABLE_SERVICES
```
jmartinez-silabs, line 92: define name isn't clear. ENABLE_SERVICES could be so many things.

examples/platform/silabs/BaseApplication.cpp
jmartinez-silabs: in addition, I am not sure this should be in an anonymous namespace. Should it not be part of the Service interface class you have in your design doc?

examples/thermostat/silabs/src/https_offload_example.cpp
jmartinez-silabs: All those example constants should have a clear comment explaining this is temporary for demo purpose

examples/thermostat/silabs/src/https_offload_example.cpp
```diff
@@ -0,0 +1,200 @@
...
+constexpr char kHttpClientUser[] = "admin";
+constexpr char kHttpClientPass[] = "admin";
+constexpr char kHttpPostData[] =
+    "employee_name=MR.REDDY&employee_id=RSXYZ123&designation=Engineer&company=SILABS&location=Hyderabad";
+
+HttpClient gHttpClient;
+
+// Binary semaphore: CMSIS mutex cannot be released from another thread; NWP/HTTP
+// completion callbacks run on the HttpClient service thread.
+osSemaphoreId_t gDoneLock = nullptr;
```
jmartinez-silabs: Needs a clearer name. What is Done? Why a get a semaphore when something is done?

examples/thermostat/silabs/src/https_offload_example.cpp
jmartinez-silabs: we could argue that since the semaphore handle is passed by a context, it could maybe not exist anymore

examples/thermostat/silabs/src/https_offload_example.cpp
```diff
@@ -0,0 +1,200 @@
...
+{
+    if (gHttpClient.IsRunning())
+    {
+        return SL_STATUS_ALREADY_INITIALIZED;
+    }
+
+    if (gDoneLock == nullptr)
+    {
+        // max_count=1, initial_count=0: Acquire blocks until completion callback Releases.
+        gDoneLock = osSemaphoreNew(1, 0, nullptr);
```
jmartinez-silabs: if max count 1, then should it be a mutex?

examples/thermostat/silabs/src/https_offload_example.cpp
```diff
@@ -0,0 +1,200 @@
...
+}
+
+CHIP_ERROR WaitForCallback(OperationWaitContext & waitCtx)
+{
+    const osStatus_t status = osSemaphoreAcquire(waitCtx.lock, osWaitForever);
+    VerifyOrReturnError(status == osOK, CHIP_ERROR_INTERNAL);
+    return waitCtx.result;
+}
+
+CHIP_ERROR RunQueuedOperation(CHIP_ERROR queueResult, OperationWaitContext & waitCtx)
```
jmartinez-silabs, line 93: I am not convinced that this implementation of `RunQueuedOperation`, `WaitForCallback` and `OperationWaitContext` is the right decision. It look overcomplicated and hard to follow for a simple semaphore/sync mechanism 

the context/semaphore is share through multiple calls but you already have 1 thread and 1 callback.

your other process should just have to wait for the unlock from the callback imo. Maybe I am misunderstanding something but there is also no comment or brief explaining what is done.

src/platform/silabs/services/http_client.cpp
```diff
@@ -0,0 +1,720 @@
...
+    if (status == SL_STATUS_ALLOCATION_FAILED)
+    {
+        return CHIP_ERROR_NO_MEMORY;
+    }
+    return CHIP_ERROR_INTERNAL;
+}
+
+bool HttpClient::IsRunning() const
+{
+    return mThreadId != nullptr;
```
jmartinez-silabs, line 64: the fact the thread exist/isn't doesn't mean it is running but ok for now

src/platform/silabs/services/http_client.cpp
```diff
@@ -0,0 +1,720 @@
...
+    VerifyOrReturnError(!mBusy, CHIP_ERROR_BUSY);
+
+    osEventFlagsSet(static_cast<osEventFlagsId_t>(mEventFlags), kEventStop);
+    while (mThreadId != nullptr)
+    {
+        osDelay(1);
+    }
+
+    osEventFlagsDelete(static_cast<osEventFlagsId_t>(mEventFlags));
+    mEventFlags = nullptr;
```
jmartinez-silabs: this cleanup could be done by the task it self before it terminates and that remove the need to do a busy wait in this calling thread

src/platform/silabs/services/http_client.cpp
```diff
@@ -0,0 +1,720 @@
...
+    {
+        CompleteOperation(CHIP_NO_ERROR);
+        return;
+    }
+
+    // Drive PUT chunking on the service thread until NWP needs another wait or upload finishes.
+    do
+    {
+        result = ProcessContinuePut();
+    } while ((result == CHIP_NO_ERROR) && mPutActive);
```
jmartinez-silabs: Can be combine in 1 while statement
```suggestion
    // Drive PUT chunking on the service thread until NWP needs another wait or upload finishes.
    while ((result == CHIP_NO_ERROR && mPutActive)
    {
        result = ProcessContinuePut();
    } 
```

examples/thermostat/silabs/src/https_offload_example.cpp
```diff
@@ -0,0 +1,200 @@
...
+{
+    if (queueResult != CHIP_NO_ERROR)
+    {
+        return queueResult;
+    }
+    return WaitForCallback(waitCtx);
+}
+
+CHIP_ERROR RunHttpsOffloadExample()
```
agarwal57, line 102: - Until this entire example is executed, the Matter thread gets blocked.

src/platform/silabs/services/http_client.h
```diff
@@ -0,0 +1,300 @@
...
+     * @param[in] body     Full request body to upload.
+     * @param[in] callback Completion callback; may be nullptr.
+     * @param[in] context  Passed to @p callback.
+     *
+     * @retval CHIP_NO_ERROR              Request queued.
+     * @retval CHIP_ERROR_BUSY            Another request is in progress.
+     * @retval CHIP_ERROR_INCORRECT_STATE Client not initialized.
+     * @retval CHIP_ERROR_INVALID_ARGUMENT Null host or resource.
+     */
+    CHIP_ERROR Put(const HttpHost & host, const char * resource, ByteSpan body, HttpOperationCallback callback,
```
agarwal57, line 204: - As all the 3 functions have same signature, and possibly similar code, we should see if we can combine these three issues into something called Request API with verb (PUT/POST/GET) as input params. --> Not to be done on this PR, we can create a JIRA for this and such other items to look into for improvement later.

src/platform/silabs/services/http_client.h
```diff
@@ -0,0 +1,300 @@
...
+                   void * context = nullptr);
+
+private:
+    static constexpr size_t kDefaultThreadStackSize = 4 * 1024;
+    // HTTP request status codes.
+    static constexpr uint8_t kHttpSuccessResponse   = 1;
+    static constexpr uint8_t kHttpFailureResponse   = 2;
+    // HTTP client error codes.
+    static constexpr uint16_t kHttpClientErrorMin   = 400;
+    static constexpr uint16_t kHttpServerErrorMax   = 599;
```
agarwal57: - Are we just using this as HttpError range only? then why not have just HttpErrorMin and Max, what is the significance of these Client and Server being present here. am aware that 4xx are client errors and 5xx are server related errors. But if we are not using those explicitly, why do we need to name like this?

src/platform/silabs/services/http_client.h
```diff
@@ -0,0 +1,300 @@
...
+        // TODO: Add Delete operation.
+        // TODO: Add Head operation.
+    };
+
+    static constexpr uint32_t kEventOperation = 1u << 0;
+    static constexpr uint32_t kEventResponse  = 1u << 1;
+    static constexpr uint32_t kEventStop      = 1u << 2;
+
+    static void ServiceThread(void * arg);
+    static CHIP_ERROR MapStatus(sl_status_t status);
```
agarwal57, line 270: - briefly describe the funcitons.

src/platform/silabs/services/http_client.h
```diff
@@ -0,0 +1,300 @@
...
+    void ProcessOperation();
+    void HandleNwpResponse();
+    CHIP_ERROR ProcessInit();
+    CHIP_ERROR ProcessDeinit();
+    CHIP_ERROR ProcessGet();
+    CHIP_ERROR ProcessPost();
+    CHIP_ERROR ProcessPut();
+    CHIP_ERROR ProcessContinuePut();
+    void CompleteOperation(CHIP_ERROR error);
+    void SignalResponse();
```
agarwal57, line 317: - please add comments for above two APIs

src/platform/silabs/services/http_client.h
```diff
@@ -0,0 +1,300 @@
...
+    Operation mPendingOperation = Operation::None;
+
+    HttpOperationCallback mUserCallback = nullptr;
+    void * mUserCallbackContext         = nullptr;
+
+    bool mInitialized = false;
+    static bool sNwpCaLoaded;
+
+    /** Client owning the request registered with the NWP; used when a callback carries no context. */
+    static HttpClient * sActiveClient;
```
agarwal57, line 351: - i thought you mentioned during discussions that you will make this class singleton. I don tthink it is according to that.

src/platform/silabs/services/http_client.cpp
```diff
@@ -0,0 +1,720 @@
...
+    mUserCallback         = callback;
+    mUserCallbackContext  = context;
+    mBusy                 = true;
+    osEventFlagsSet(static_cast<osEventFlagsId_t>(mEventFlags), kEventOperation);
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR HttpClient::ProcessContinuePut()
+{
+    if (mEndOfFile == kHttpSuccessResponse)
```
agarwal57: - Why is the mEndOfFile is being assigned or being validated with the kHttpSuccessResponse? If it is 0 or 1, then we should just use the flag accordingly rather than assigning a value which doesnt seem to relevant to the member variable.

src/platform/silabs/services/http_client.cpp
```diff
@@ -0,0 +1,720 @@
...
+    }
+
+    const int32_t totalPutDataLen = static_cast<int32_t>(mPutBody.size());
+    mHttpChunkLength              = ((totalPutDataLen - mHttpOffset) > SL_HTTP_CLIENT_MAX_WRITE_BUFFER_LENGTH)
+                     ? SL_HTTP_CLIENT_MAX_WRITE_BUFFER_LENGTH
+                     : (totalPutDataLen - mHttpOffset);
+
+    if (mHttpChunkLength > 0)
+    {
+        sl_status_t status = sl_http_client_write_chunked_data(&mClientHandle, const_cast<uint8_t *>(mPutBody.data() + mHttpOffset),
```
agarwal57: - This is mentioned only for PUT, not for POST? POST can also have a long body right?

src/platform/silabs/services/http_client.cpp
```diff
@@ -0,0 +1,720 @@
...
+        }
+    }
+
+    if (mCredentials != nullptr)
+    {
+        free(mCredentials);
+        mCredentials = nullptr;
+    }
+
+    mResponseBytesWritten = 0;
```
agarwal57: - Already most of these flags are reset in resetRequestState function. We can call that instead of duplicating.

src/platform/silabs/services/http_client.cpp
```diff
@@ -0,0 +1,720 @@
...
+    {
+        sActiveClient = nullptr;
+    }
+    ChipLogProgress(DeviceLayer, "HTTPS client deinit success");
+    return CHIP_NO_ERROR;
+}
+
+void HttpClient::CompleteOperation(CHIP_ERROR error)
+{
+    if (error != CHIP_NO_ERROR)
```
agarwal57, line 379: - Add comments here.

src/platform/silabs/services/http_client.cpp
```diff
@@ -0,0 +1,720 @@
...
+    if ((getResponse->http_response_code >= kHttpClientErrorMin) && (getResponse->http_response_code <= kHttpServerErrorMax) &&
+        (getResponse->http_response_code != 0))
+    {
+        self->mHttpRspReceived = kHttpFailureResponse;
+        self->mCallbackStatus  = SL_STATUS_FAIL;
+        self->SignalResponse();
+        return getResponse->status;
+    }
+
+    if (getResponse->data_length)
```
agarwal57, line 472: - What happens when the buffer is smaller than the response of the get request.

src/platform/silabs/services/http_client.cpp
```diff
@@ -0,0 +1,720 @@
...
+
+    if ((postResponse->status != SL_STATUS_OK) && (postResponse->status != SL_STATUS_IN_PROGRESS))
+    {
+        self->mHttpRspReceived = kHttpFailureResponse;
+        self->SignalResponse();
+        return postResponse->status;
+    }
+
+    if ((postResponse->http_response_code >= kHttpClientErrorMin) && (postResponse->http_response_code <= kHttpServerErrorMax) &&
+        (postResponse->http_response_code != 0))
```
agarwal57: - What is 0 means and add comments here

src/platform/silabs/services/http_client.cpp
```diff
@@ -0,0 +1,720 @@
...
+    if (postResponse->end_of_data)
+    {
+        self->mHttpRspReceived = kHttpSuccessResponse;
+        self->SignalResponse();
+    }
+
+    return SL_STATUS_OK;
+}
+
+sl_status_t HttpClient::PutResponseCallback(const sl_http_client_t * client, sl_http_client_event_t event, void * data,
```
agarwal57, line 548: - similar to the request, we can plan to combine the callbacks as well. --> Again not for this PR, we can handle it separately.

agarwal57 (changes requested):

examples/platform/silabs/BaseApplication.cpp
rosahay-silabs: > This is a common base for all examples, but `http_client_demo_start` is both declared and defined in `thermostat/silabs`. Other examples cannot be compiled with SL_MATTER_ENABLE_HTTP_SERVICE. Either `http_client_demo_start` must be called from the thermostat init, or the function should be part of the common code.

I agree, for the `http_client_demo_start` to work from application task context, I have to pass the WiFi and got IP indication to the application, this code will be moved into application once that is done.

examples/platform/silabs/BaseApplication.cpp
rosahay-silabs: `SL_MATTER_ENABLE_SERVICES` was introduced to enable common functionality that can be leveraged across multiple services, such as the `startAll`/`stopAll` functions. Any functionality specific to an individual service is enabled separately under its respective service-specific macro.

examples/thermostat/silabs/certs/cacert.pem.h
rosahay-silabs: @rcasallas-silabs , it is used by both HTTPs client and MQTTs client, I will rename it properly though but this will ultimately will be a blank file on the final commit, to customer.

examples/thermostat/silabs/src/https_offload_example.cpp
rosahay-silabs: Let me explain, the flow in the demo is that the `index.html` is uploaded to the HTTPs using `PUT` and then fetched using `GET` and then a value is posted to the server on `index.html`. Hence unless `PUT` is successful and completed, `GET` would typically not get the resource on the particular URI and fail. This is purely part of how the WiseConnect example was but since it was a single thread, the `PUT` operations were done completely in a synchronous manner, before invoking `GET`.

examples/platform/silabs/BaseApplication.cpp
rosahay-silabs: In this context just the demo code will fail to run/execute if the timer fails, but yes timer failing is a bigger concern which is not a part of scope of the PR.
I will align it to `TEMPORARY_RETURN_IGNORED`, similar to the rest of code.

examples/thermostat/silabs/src/https_offload_example.cpp
rosahay-silabs: fixed

examples/thermostat/silabs/src/https_offload_example.cpp
rosahay-silabs: ah yes it typically should be, but since I am trying to release it in the context of a different thread as in the callback, hence the use of semaphore.

> // Binary semaphore: CMSIS mutex cannot be released from another thread; NWP/HTTP
// completion callbacks run on the HttpClient service thread.

examples/thermostat/silabs/src/https_offload_example.cpp
rosahay-silabs: done

examples/thermostat/silabs/include/https_offload_example.h
rosahay-silabs: done, but since the same component can be used for HTTP also by enabling the `httpsEnable` flag while init, I have kept the name of the class as `HttpClient` and not `HttpsClient` meanwhile renamed the instance to `gHttpsClient` instead of `gHttpClient`, to maintain clarity that this object is to be configured and used for HTTPs operations.

examples/platform/silabs/BaseApplication.cpp
rosahay-silabs: Since failing to schedule does mean something more is wrong with the code that just the part we are adding here, I am aligning it to `TEMPORARY_RETURN_IGNORED` as per the rest of the implementation, for now.

examples/thermostat/silabs/certs/cacert.pem.h
rosahay-silabs: Renamed it to `kCaCertExample`

examples/thermostat/silabs/src/https_offload_example.cpp
rosahay-silabs: Added a comment indicating the same, also removed the hardcoded values

examples/thermostat/silabs/src/https_offload_example.cpp
rosahay-silabs: Added a `@breif` to the file

src/platform/silabs/services/http_client.cpp
rosahay-silabs: right! moved it to the proper location

src/platform/silabs/services/http_client.cpp
rosahay-silabs: Yes, correct making changes to get in a generic `ProcessContinueChunkedBody` instead of `ProcessContinuePut` which will be more generic to both of the items

src/platform/silabs/services/http_client.cpp
rosahay-silabs: yes! added `ResetRequestState` instead.

src/platform/silabs/services/http_client.cpp
rosahay-silabs: Only copy over the amount of data the provided buffer allows, rest drop the remaining bytes with an error log

examples/thermostat/silabs/src/https_offload_example.cpp
rosahay-silabs: yes, true for now, I will move it to the application context in the next PR, maybe on button press or after programming the indications to the application layer and hook this example to `IP_GOT` event instead.

examples/thermostat/silabs/src/https_offload_example.cpp
rosahay-silabs: Let me check if I can simplify this example.

Overall, the dependency here is that the example uses a chunked `PUT` request to upload the `index.html` file. Until the upload is complete, invoking a `GET` request for the same `index.html` resource will return an appropriate error indicating that the client is busy processing the ongoing `PUT` request.
So technically, the `GET` API is behaving as expected by returning the busy error. However, we would still need to test the actual `GET` functionality once the `PUT` operation is complete.

src/platform/silabs/services/http_client.cpp
rosahay-silabs: just for my information, what would be the way to be sure? I thought keeping the `threadId` should be indicative of the thread.

src/platform/silabs/services/http_client.cpp
rosahay-silabs: This can be chunked too, but I will explore this in the later PR. Since I will have to test and check with sample application first on the behaviour that is documented.

src/platform/silabs/services/http_client.cpp
rosahay-silabs: will add this to a JIRA task

src/platform/silabs/services/http_client.cpp
rosahay-silabs: aligned it with the `sl_http_client.h` instead of the sample usage of the flag

```cpp
  uint32_t end_of_data; ///< Indicates the completion status of a data chunk or communication phase.
    ///<
    ///< GET/POST (indications from server):
    ///<   0: More response data is expected.
    ///<   1: This is the final chunk of the response.
    ///<
    ///< PUT:
    ///<   Acknowledgment for transmitted data (notification from NWP):
    ///<     0: Further data transmission is expected.
    ///<     1: Data transmission is complete.
    ///<   Server response (data received from server after transmission):
    ///<     8: More response data is expected.
    ///<     9: This is the final chunk of the response. See @ref SL_HTTP_CLIENT_PUT_SERVER_RESPONSE_END_OF_DATA.
```

src/platform/silabs/services/http_client.h
```diff
@@ -0,0 +1,301 @@
...
+struct HttpClientConfig
+{
+    uint8_t certificateIndex = 1;       ///< NWP TLS / HTTPS certificate index.
+    bool httpsEnable         = true;    ///< true = HTTPS, false = plain HTTP.
+    const char * username    = nullptr; ///< Optional HTTP basic-auth username; nullptr skips auth.
+    const char * password    = nullptr; ///< Optional HTTP basic-auth password; nullptr skips auth.
+
+    /** PEM bytes for sl_net_set_credential(SL_NET_SIGNING_CERTIFICATE). nullptr = skip CA load. */
+    const uint8_t * tlsCaCert = nullptr;
+    size_t tlsCaCertLen       = 0; ///< Byte length of @p tlsCaCert (exclude trailing NUL for string PEMs).
```
sabollim-silabs, line 57: do we need this as config or derive it from key itself ?

src/platform/silabs/services/http_client.cpp
rosahay-silabs: done

src/platform/silabs/services/http_client.h
```diff
@@ -0,0 +1,301 @@
...
+/**
+ * @brief Per-request destination for Get / Post / Put.
+ *
+ * The SiWx917 NWP HTTP client needs a connect address (@p serverIp) and a Host / SNI
+ * name (@p hostName) separately; they may be the same string.
+ */
+struct HttpHost
+{
+    const char * hostName = nullptr; ///< Host header / SNI name. Must not be nullptr for requests.
+    const char * serverIp = nullptr; ///< IP (or resolvable address) used to connect. Must not be nullptr.
```
sabollim-silabs, line 69: should we provide the option for URL and DNS lookup to get ip from it

src/platform/silabs/services/http_client.cpp
rosahay-silabs: though there is not much documentation on this, I am using `SL_STATUS_OK` instead of `0`
NOTE: add this to follow up JIRA for additional clarification on the `http_response_code` expected outcomes and why `kHttpClientErrorMin` `kHttpServerErrorMax` range was shared on the example.

src/platform/silabs/services/http_client.h
rosahay-silabs: The naming is aligned with the minimum HTTP Client code which starts from 400 and HTTP Server code which reaches up 599 in the 5XX range, this is used to filter out the other error codes, I will check with WiseConnect if there was a reason it was programmed that way in the example and if we have some limitations on this.

src/platform/silabs/services/http_client.h
rosahay-silabs: done

src/platform/silabs/services/http_client.h
rosahay-silabs: done

src/platform/silabs/services/http_client.h
rosahay-silabs: As discussed offline, we can take this up in the next PR based on the better pros and cons.

src/platform/silabs/services/http_client.h
rosahay-silabs: Since we are expecting it as `uint8_t *` aka data, thus not expected to run `strnlen` to fetch the length. Also we were to run `sizeof` the pointer will just return the size of pointer type based on the compiler. Hence that will be incorrect, let me know if I missing something here.

src/platform/silabs/services/http_client.h
rosahay-silabs: I need further examples to check how is this supported on the HTTPs client, post that download I can verify on how to add this to our client code.

src/platform/silabs/services/http_client.cpp
jmartinez-silabs: Depends what you want to consider "running" but osThreadGetState() will get you the exact state and then you can define what you care as running/non running
https://arm-software.github.io/CMSIS_5/RTOS2/html/group__CMSIS__RTOS__ThreadMgmt.html#gad3dc89e942e38d9f3af858a0269a820d

examples/thermostat/silabs/certs/cacert.h
```diff
@@ -0,0 +1,24 @@
+/*
```
jepenven-silabs, line 1: this file extension seems kind of weird with the double `.pem.h` which can be misleading since `.pem` is a valid extension for certificate....

examples/thermostat/silabs/resources/index.html.h
```diff
@@ -0,0 +1,39 @@
...
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+/*
+ * @file  index.html.h
+ * @brief HTTP page for the HTTPs offload client application,
+ * this page is uploaded to the server.
+ */
+
```
jepenven-silabs: 2 problems, double extension (again ?) missing header guards

examples/thermostat/silabs/certs/cacert.h
```diff
@@ -0,0 +1,24 @@
...
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+/**
+ * @file  cacert.pem.h
+ * @brief Sample (generated) CA certificate for HTTPs and MQTTs clients
+ */
+
```
jepenven-silabs, line 23: missing header guards

examples/thermostat/silabs/src/https_offload_example.cpp
```diff
@@ -0,0 +1,220 @@
...
+ * 5. Run the HTTPS GET request to get the index.html file from the server
+ * 6. Run the HTTPS POST request to send the post data to the server
+ * 7. Complete the HTTPS offload example and return the result
+ * 8. Deinitialize the HTTPS client (optional, disabled)
+ * 9. Stop the HTTPS client (optional, disabled)
+ */
+
+#include "https_offload_example.h"
+
+#include "cacert.pem.h"
```
jepenven-silabs: yeah the double `. .` is not friendly with all compilers...

examples/thermostat/silabs/src/https_offload_example.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+// TODO: Update the below values to the correct values for your server.
+constexpr char kHttpServerIp[]   = "";
+constexpr char kHttpHostname[]   = ""; // used for SNI validation
+constexpr uint16_t kHttpPort     = 8443;
+constexpr char kHttpUrl[]        = "/index.html";
+constexpr char kHttpClientUser[] = "admin";
+constexpr char kHttpClientPass[] = "admin";
+constexpr char kHttpPostData[] =
+    "employee_name=MR.REDDY&employee_id=RSXYZ123&designation=Engineer&company=SILABS&location=Hyderabad";
```
jepenven-silabs: As far as demo/default value goes, I've seen more generic ones `John Doe` `foo bar` etc...

examples/thermostat/silabs/src/https_offload_example.cpp
jepenven-silabs: Even with this explanation @jmartinez-silabs comments is still valid. `gHttpsStackLock` / `gHttpsTaskLock` would be more explicit than `gDoneLock`

examples/thermostat/silabs/src/https_offload_example.cpp
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    const HttpClientConfig config = {
+        .certificateIndex = 1,
+        // Enable HTTPS client.
+        .httpsEnable  = true,
+        .username     = kHttpClientUser,
+        .password     = kHttpClientPass,
+        .tlsCaCert    = reinterpret_cast<const uint8_t *>(kCaCertExample),
+        .tlsCaCertLen = sizeof(kCaCertExample) - 1,
```
jepenven-silabs, line 176: This seems wrong. Since is defined out of scope what would happen if the sizeof kCaCertExample is 0 ? I know it "shouldn't" happen but a little check here wouldn't hurt

examples/thermostat/silabs/src/https_offload_example.cpp
```diff
@@ -0,0 +1,220 @@
...
+    }
+    ChipLogProgress(DeviceLayer, "HTTPS POST request success");
+
+    ChipLogProgress(DeviceLayer, "HTTPS demo completed");
+    return CHIP_NO_ERROR;
+}
+
+} // namespace
+
+sl_status_t https_client_demo_start(void)
```
jepenven-silabs, line 149: why the usage of `sl_status_t` since most of the API used in this function returns CHIP_ERROR?

examples/thermostat/silabs/src/https_offload_example.cpp
```diff
@@ -0,0 +1,220 @@
...
+    };
+
+    CHIP_ERROR err = gHttpsClient.Start();
+    if (err != CHIP_NO_ERROR)
+    {
+        ChipLogError(DeviceLayer, "HTTPS Start failed: %" CHIP_ERROR_FORMAT, err.Format());
+        return SL_STATUS_FAIL;
+    }
+
+    OperationWaitContext waitCtx = { .lock = gDoneLock, .result = CHIP_NO_ERROR };
```
jepenven-silabs: Considering that all `OperationWaitContext` uses the same semaphore, this whole logic seems like bloats that increase RAM/Stack usage for little to no gain....

src/platform/silabs/services/matter_service.h
```diff
@@ -0,0 +1,69 @@
...
+
+namespace chip {
+namespace DeviceLayer {
+namespace Silabs {
+
+namespace {
+/**
+ * @brief Delay in seconds before initializing the Matter services.
+ */
+static constexpr uint32_t kMatterServicesInitDelaySec = 10;
```
jepenven-silabs, line 31: a little more details on why this need to be 10 sec would be nice

jepenven-silabs (commented):
beyond the header guards/size check, my other comments can be fixed in a follow up PR.

jepenven-silabs (approved):

sabollim-silabs (approved):

examples/thermostat/silabs/src/https_offload_example.cpp
jmartinez-silabs: I'm still highly sceptical of the approach. If you want to post pone this, create a task on JIRA for the phase

examples/thermostat/silabs/src/https_offload_example.cpp
jmartinez-silabs: Would be great to have this configurable in come way that doesn't require modifying a c file and recompiling. 

Even just some define, that can be overwritten, setting those constant would be an improvement

examples/platform/silabs/BaseApplication.cpp
rosahay-silabs: For now it is part of the `.slcc` but will put it as part of configuration then, I have to look into that part.

examples/thermostat/silabs/certs/cacert.pem.h
```diff
@@ -0,0 +1,24 @@
+/*
```
rosahay-silabs: done

examples/thermostat/silabs/certs/cacert.pem.h
```diff
@@ -0,0 +1,24 @@
...
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+/**
+ * @file  cacert.pem.h
+ * @brief Sample (generated) CA certificate for HTTPs and MQTTs clients
+ */
+
```
rosahay-silabs: done

examples/thermostat/silabs/resources/index.html.h
rosahay-silabs: done

examples/thermostat/silabs/src/https_offload_example.cpp
rosahay-silabs: done changes

examples/thermostat/silabs/src/https_offload_example.cpp
rosahay-silabs: done

examples/thermostat/silabs/src/https_offload_example.cpp
rosahay-silabs: done

examples/thermostat/silabs/src/https_offload_example.cpp
rosahay-silabs: sure, added a check

examples/thermostat/silabs/src/https_offload_example.cpp
rosahay-silabs: I will comeback to this and simplify the example application, it does look more convoluted than needed.

src/platform/silabs/services/http_client.h
rosahay-silabs: Yes, it was 2kB with the previous approach where we did not have a client code, now with the client code in between, the stack depth increased, I ballparked it to 4kB, but 3kB also might work.

PR #1153, opened 2026-08-18
[SL-ONLY] Update wake-on-matter

jepenven-silabs:
### Summary

PR that updates the configuration and general behavior of the wake on Matter app 
- Remove dependency for BaseApplication (useless in the NCP context)
- Change the Init for single thread instance
- Add logic for GPIO toggle upon subscription reception

### Dependencies 
this PR will require this PR before merging 

### Related issues
#MATTER-6390
### Testing
Tested locally


examples/wake-on-matter/silabs/src/AppTask.cpp
```diff
@@ -270,6 +276,87 @@ void UnregisterRootNodeClusters(CodeDrivenDataModelProvider & provider)
     }
 }
 
+// GPIO used to wake up the host. For now routed to the LED1 on the MG24 BRD4187c
+// Adjust port/pin as needed.
+constexpr sl_gpio_t kLightGpio         = { .port = gpioPortB, .pin = 4 };
```
jmartinez-silabs, line 281: should it be named like WakeUpGpio or something like that

examples/wake-on-matter/silabs/src/AppTask.cpp
```diff
@@ -270,6 +276,87 @@ void UnregisterRootNodeClusters(CodeDrivenDataModelProvider & provider)
...
+constexpr uint16_t kLedAutoOffTimeoutS = 5;
+
+// Matter cluster / attribute identifiers we react to.
+constexpr uint32_t kOnOffClusterId              = 0x00000006;
+constexpr uint32_t kOnOffAttributeId            = 0x00000000;
+constexpr uint32_t kOccupancySensingClusterId   = 0x00000406;
+constexpr uint32_t kOccupancyAttributeId        = 0x00000000;
+constexpr uint32_t kBooleanStateClusterId       = 0x00000045;
+constexpr uint32_t kBooleanStateValueAttributeId = 0x00000000;
```
jmartinez-silabs, line 290: Could those be taken from the metadata e.g
zzz_generated/app-common/clusters/OnOff/ClusterId.h

examples/wake-on-matter/silabs/src/AppTask.cpp
```diff
@@ -296,64 +383,110 @@ DataModel::Provider * AppTask::GetDataModelProvider()
...
-
 static otInstance * sInstance = NULL;
+extern "C" otInstance * otInstanceInitSingle(void) {}
 #endif
-
 extern "C" void sl_ot_ncp_init(void)
 {
+    if(otGetInstance() == nullptr)
+    {
+        sl_ot_create_instance();
```
jmartinez-silabs, line 481: How does that interact with our ThreadStackManagerImpl. Is there come build config/option that makes it so we don't use our typical matter threadStackManager?

Same question apply to the `otSetStateChangedCallback` below

examples/wake-on-matter/silabs/src/AppTask.cpp
```diff
@@ -270,6 +276,87 @@ void UnregisterRootNodeClusters(CodeDrivenDataModelProvider & provider)
...
+constexpr uint32_t kOnOffAttributeId            = 0x00000000;
+constexpr uint32_t kOccupancySensingClusterId   = 0x00000406;
+constexpr uint32_t kOccupancyAttributeId        = 0x00000000;
+constexpr uint32_t kBooleanStateClusterId       = 0x00000045;
+constexpr uint32_t kBooleanStateValueAttributeId = 0x00000000;
+
+// Occupancy attribute is bitmap8; bit 0 set means the sensor reports "occupied".
+constexpr uint64_t kOccupancyOccupiedMask = 0x01;
+
+void EnsureLightGpioInitialized()
```
jmartinez-silabs, line 295: not the end of the world, but i'm questioning the use of `Light` in all those api

jmartinez-silabs (approved):

examples/wake-on-matter/silabs/README.md
```diff
@@ -171,9 +171,177 @@ Steps
 ```
     chip-tool accesscontrol write acl '[{"fabricIndex": 1, "privilege": 1, "authMode": 2, "subjects": [<woM node ID>], "targets": null}]' <target node id> 0
 
+Important detail
+write acl replaces the full ACL list, it does not append one entry. So if you already have other ACL entries, you must include them too in the same JSON array, for example:
```
malbert-silabs, line 175: This text seems to be inside the \`\`\` code block, missing a  \`\`\` before "Important detail"
Also maybe add a formatting to "Important detail" to add importance? Like bolding or a # header?

examples/wake-on-matter/silabs/README.md
```diff
@@ -171,9 +171,177 @@ Steps
...
+- `scripts/tests/end_to_end_validation.sh` - runs the checks above, then
+  commissions an accessory + the wake-on-matter NCP, writes the ACL, and
+  establishes a subscription.
+
+## Prerequisites
+
+Host packages (Ubuntu):
+
+```sh
+sudo apt-get install net-tools     # provides ifconfig, required by verify_install.sh
```
malbert-silabs, line 206: nit but maybe `apt` is better than `apt-get` in this context?

malbert-silabs (approved):
LGTM, approving on comment resolution

examples/wake-on-matter/silabs/src/AppTask.cpp
```diff
@@ -296,64 +383,110 @@ DataModel::Provider * AppTask::GetDataModelProvider()
...
-        appError(err);
+        SILABS_LOG("SetThreadProvision failed: %" CHIP_ERROR_FORMAT, err.Format());
+        return;
     }
 
-    SILABS_LOG("App Task started");
+    err = ThreadStackMgr().SetThreadEnabled(true);
+    if (err != CHIP_NO_ERROR)
+    {
+        SILABS_LOG("SetThreadEnabled failed: %" CHIP_ERROR_FORMAT, err.Format());
```
Sarthak-Shaha, line 448: Do we want to move to use chiplog* macros instead of silabs_log ?

Sarthak-Shaha (approved):

examples/wake-on-matter/silabs/src/AppTask.cpp
jepenven-silabs: So actually we do use it, the `otGetInstance` routes to our `ThreadStackManager` and same thing with `sl_ot_create_instance` 

The `if` is actually a safeguard to prevent an init race condition if we don't know who will initialize the instance first.

examples/wake-on-matter/silabs/src/AppTask.cpp
jepenven-silabs: yes, I'm waiting for PR to get merge and the trigger point for waking up will be dynamic so that they can be injected by the host application

examples/wake-on-matter/silabs/src/AppTask.cpp
jepenven-silabs: a reasonnable question indeed. The API will be renamed in a follow up PR one the Host application glue is merged.

examples/wake-on-matter/silabs/src/AppTask.cpp
jepenven-silabs: will change in a follow-up

examples/wake-on-matter/silabs/README.md
jepenven-silabs: will change in a follow-up PR

examples/wake-on-matter/silabs/README.md
jepenven-silabs: will change in a follow-up PR (net-tools dependency might drop)

examples/wake-on-matter/silabs/src/AppTask.cpp
jepenven-silabs: might want this is debug code might be removed later on

PR #1156, opened 2026-08-19
[CSA-CP] Fix LCD Update after Commissioning CSA #73440

korino-silabs:
##### Problem

- After commissioning, the Silabs window-app LCD did not refresh to the window-covering UI.
- Cover painting only ran from `UpdateLCD()` on attribute/UI state changes, and commissioning does not produce those changes.
- Unlike other Silabs examples, window-app did not register a custom DemoScreen UI, so the shared post-commissioning `DemoScreen` update path did not draw the cover.

##### Solution

- Register `WindowManager::DrawUI` via `SetCustomUI()` and initialize the LCD with `WriteDemoUI()` / QR display when unprovisioned, matching other Silabs apps (e.g. lock).
- This allows `BaseApplication`'s commissioning-window-closed `DemoScreen` update to invoke the window cover UI.

#### Testing

- Tested on BRD4338A SoC 
- Commissioned the device repeatedly and confirmed LCD switches from QR to window-covering UI after commissioning completes.
- Power-cycled a provisioned device and confirmed cover UI is shown at boot.
- Issued Window Covering cluster commands and confirmed LCD updates on position/state changes.

### DIFF
--- a/examples/window-app/silabs/src/AppTask.cpp
+++ b/examples/window-app/silabs/src/AppTask.cpp
@@ -59,7 +59,7 @@ CHIP_ERROR AppTask::AppInit()
 	return err;
     }
 
-#if SL_MATTER_DISPLAY_ENABLED
+#if DISPLAY_ENABLED
     GetLCD().SetCustomUI(WindowManager::DrawUI);
     GetLCD().WriteDemoUI(false);
 #if SL_MATTER_QR_CODE_ENABLED
@@ -68,7 +68,7 @@ CHIP_ERROR AppTask::AppInit()
         GetLCD().ShowQRCode(true);
     }
 #endif // SL_MATTER_QR_CODE_ENABLED
-#endif // SL_MATTER_DISPLAY_ENABLED
+#endif // DISPLAY_ENABLED
 
     return CHIP_NO_ERROR;
 }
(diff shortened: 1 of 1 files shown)


chirag-silabs (approved):

chirag-silabs:
just check, if it can be pulled to the v1.6.1 branch directly rather than a cherry-pick

examples/window-app/silabs/src/AppTask.cpp
```diff
@@ -56,9 +56,21 @@ CHIP_ERROR AppTask::AppInit()
     {
         SILABS_LOG("WindowMgr::Init() failed");
         appError(err);
+	return err;
     }
 
-    return err;
+#if DISPLAY_ENABLED
```
Sarthak-Shaha, line 62: Should it be #ifdef to match previous declarations?

korino-silabs: yes, it is supposed to be checking ifdef, I accidentally created this PR, the updated PR is https://github.com/project-chip/connectedhomeip/pull/73643 which would sync up onto 2.10

The updated PR contains the proper handling

Can i close this PR?

Sarthak-Shaha: yes, please close.

korino-silabs: sure!

korino-silabs:
Changes would sync up from https://github.com/project-chip/connectedhomeip/pull/73643

PR #1159, opened 2026-08-20
[SL-ONLY] Fix BLE dependency bug in gn

jepenven-silabs:
### Summary
BLE shell command are not gated on the presence of the BLE manager.

### Related issues

None

### Testing
CI should build without any issue

### DIFF
--- a/examples/platform/silabs/efr32/BUILD.gn
+++ b/examples/platform/silabs/efr32/BUILD.gn
@@ -226,7 +226,9 @@ source_set("efr32-common") {
     if (matter_enable_tracing_support) {
       public_deps += [ "${silabs_common_plat_dir}/shell:tracing" ]
     }
-    deps += [ "${silabs_common_plat_dir}/shell:ble" ]
+    if (chip_config_network_layer_ble) {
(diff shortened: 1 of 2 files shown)


Sarthak-Shaha (approved):

examples/platform/silabs/efr32/BUILD.gn
```diff
@@ -226,7 +226,9 @@ source_set("efr32-common") {
     if (matter_enable_tracing_support) {
       public_deps += [ "${silabs_common_plat_dir}/shell:tracing" ]
     }
-    deps += [ "${silabs_common_plat_dir}/shell:ble" ]
+    if (chip_config_network_layer_ble) {
```
Sarthak-Shaha, line 229: Always wondered if we can move the commonly used gn configs to the common file.

jepenven-silabs: usually yes. a cleanup could be done on that side but since our main offering is now SLC based, it was never prioritized

jmartinez-silabs (approved):

